PR #7, opened 2020-01-14
Phase out as_port_learn and as_port_status

anurag6:


forch/faucet_event_client.py
```diff
@@ -215,8 +214,16 @@ def next_event(self, blocking=False):
             except Exception as e:
                 LOGGER.info('Error (%s) parsing\n%s*\nwith\n%s*', str(e), line, remainder)
                 continue
-            if self._filter_faucet_event(event):
-                if not self._dispatch_faucet_event(event):
+            event_target = None
+            target_event = None
+            for target in self._handlers:
```
didovesei: Seems that this for loop does not produce any results? Is it an indenting error?

anurag6: It sets target_event and event_target if the target exists in the list of targets we have. They are then used in _filter_faucet_event and _dispatch_faucet_event

forch/faucet_event_client.py
```diff
@@ -193,15 +196,11 @@ def set_event_horizon(self, event_horizon):
...
-    def _dispatch_faucet_event(self, event):
-        for target in self._handlers:
-            if target in event:
-                faucet_event = dict_proto(event, FaucetEvent, ignore_unknown_fields=True)
-                target_event = getattr(faucet_event, target)
-                self._augment_event_proto(faucet_event, target_event)
-                LOGGER.debug('dispatching %s event', target)
-                self._handlers[target](target_event)
-                return True
+    def _dispatch_faucet_event(self, target_event, target):
```
grafnu: switch order of arguments -- should be "target" and "target_event"

forch/faucet_event_client.py
grafnu: in general this coding style is problematic b/c it's confusing.  Better to accumulate all results in a list, and then explicitly use [0] from that list. You can/should also use a list comprehension for it, as it's much cleaner overall. 

Also, should minimize the loop (or comprehension) to the minimum necessary, which is selecting the matching "target" value -- that's all that really needs to happen, and then everything else is based on that.

forch/faucet_event_client.py
```diff
@@ -105,23 +105,26 @@ def has_event(self, blocking=False):
             else:
                 return False
 
-    def _filter_faucet_event(self, event):
+    def _filter_faucet_event(self, event, target_event):
```
grafnu: this method needs some work since it's either misnamed or doing too many different things.  It's now taking two events, which really muddles the picture. Also, the parameters are confusing.  I would take a look at it and maybe break it out into the necessary component parts.  E.g. one that's managing the event sequence checking, another for the debounce logic.

anurag6: could the new functions still be called from within the filter_faucet_event method. It uses it as a check for whether to dispatch the event or not.

anurag6: Or I could check each of the new methods. Fair.

anurag6: Leaving in event order and target field check.

anurag6:
@grafnu hopefully the tests go through. PTAL.

forch/faucet_event_client.py
grafnu: I think you can simplify by removing target_event here, and doing the check where it's called... -- can it just be

dispatch = self._check_event_order(event) and target_event

?  Seems like this method is unnecessary.

forch/faucet_event_client.py
```diff
@@ -105,23 +105,37 @@ def has_event(self, blocking=False):
...
+        if not self._check_event_order(event):
+            return False
+        if not target_event:
             return False
-        self._last_event_id += 1
-        if event_id != self._last_event_id:
-            raise Exception('Out-of-sequence event id #%d' % event_id)
+        return True
+
+    def _check_event_order(self, event):
```
grafnu: _valid_event_order?   Give some indication of the meaning of the return

forch/faucet_event_client.py
```diff
@@ -105,23 +105,37 @@ def has_event(self, blocking=False):
...
+        event_id = int(event.get('event_id'))
+        if not event.get('debounced'):
```
grafnu: if event.get('debounced'):
  return True

Less indentation is better.  Early exit is better.

event_id check can be moved after debounce check, and should remove .get (because then event_id should always be defined, right?).

anurag6:
Addressed comments. PTAL.

anurag6:
All checks passed, merging in.

PR #8, opened 2020-01-14
use get() to get port

didovesei:


forch/faucet_state_collector.py
```diff
@@ -1006,7 +1006,7 @@ def _get_port_attributes(self, switch, port):
 
     def _get_egress_port(self, switch):
         """Get egress port of a switch"""
-        for port in self.switch_states[switch][PORTS]:
+        for port in self.switch_states[switch].get(PORTS, {}):
```
grafnu: underlying problem here is how/when is a switch defined (entry in switch_states) but has no ports.  Just something to think about for future designs... the system is somehow in a logical inconsistency which is manifesting as an error.

PR #13, opened 2020-01-22
Set of example faucet configuration and ACL files

didovesei:


didovesei:
> This should go in a parallel directory structure, rather than side by side. So, e.g., there is currently topo/bond/faucet.yaml -- but then rather than this faucet_base.yaml, it would be topo/dva/faucet.yaml -- so then one could run "bin/setup_stack bond" to get one configuraiton, and "bin/setup_stack dva" to get another (this) configuration. That would cover the "base" static structural configs (which should just be the open default vlan).
> 
> The other files, then would have a regular naming structure that represent the various dynamic test modes. So, faucet.yaml would be the (default) file, and then rather than "topology-all0w-vlan-300.yaml" -- it would be "faucet_allow300.yaml" or something -- so that dynamically a testing setup could do something like "dynamic_configure allow300" and end up doing the right thing (setup the system for testing that mode).
> 
> So, all the same content you have, just structured differently in the filesystem.

- Moved .yaml files to topo/dva
- Added etc/corp_dva.yaml based on etc/corp_nobond.yaml

PR #14, opened 2020-01-22
Faucetizer class

didovesei:


didovesei:
> High level comments:
> 
> 1. No queue or threading in this class (you already know this)
Removed

> 2. Reorganize test files so they're in topo/dva or similar (not topo/bond) -- as per other PR
Done

> 3. Add a simple test -- e.g. running the **main** you have from bin/test_topology
Added

forch/faucetizer.py
```diff
@@ -1,49 +1,164 @@
...
-        sorted_rules = list(self._network_state.get("named_match_rules", {}).keys())
-        sorted_rules.sort()
-        sys.stdout.write(f'{sorted_rules}\n')
+        self._faucetize(macs)
 
-        sorted_macs = list(self._network_state.get('device_mac_behaviors', {}).keys())
-        sorted_macs.sort()
-        sys.stdout.write(f'{sorted_macs}\n')
+    def _process_device_learning(self, mac, learning):
+        device_map = self._devices.setdefault(mac, {})
```
grafnu: the return result of this is not a "device_map" -- it's a device_entry

forch/faucetizer.py
```diff
@@ -1,49 +1,164 @@
...
+        device_map['new_switch'] = learning.switch
+        device_map['new_port'] = learning.port
```
grafnu: why convert to named dict entries -- why not just keep the underlying proto buf object?

dicts with string keys is more fragile, so unless there's a reason it's best to avoid

forch/faucetizer.py
```diff
@@ -1,49 +1,164 @@
...
 
+    def _process_device_behavior(self, mac, behavior):
+        device_map = self._devices.setdefault(mac, {})
```
grafnu: ditto to above

forch/faucetizer.py
```diff
@@ -1,49 +1,164 @@
...
+        if not self._faucet_config:
+            return
+
+        if not macs:
+            macs = self._devices.keys()
+
+        commit = False
+
+        for mac in macs:
+            old_switch = self._devices.get(mac, {}).get('old_switch')
```
grafnu: replace with attributes not string keys

if necessary you can create a protobuf message for whatever this object is

forch/faucetizer.py
```diff
@@ -1,49 +1,164 @@
...
+
+            if new_switch and vlan and role:
+                switch_config = self._faucet_config.get('dps', {}).get(new_switch, {})
+                port_config = switch_config.get('interfaces', {}).get(new_port)
+                if not port_config:
+                    LOGGER.warning('Switch or port not defined in faucet config: %s %s',
+                                   new_switch, new_port)
+                    continue
+                port_config['native_vlan'] = vlan
+                port_config['acls_in'] = [f'role_{role}']
```
grafnu: what if role is not specified?

bin/test_topology
```diff
@@ -7,14 +7,28 @@ preamble_setup
...
 if [ -z "$VIRTUAL_ENV" -a -f venv/bin/python3 ]; then
     PYTHON=venv/bin/python3
 else
     PYTHON=python3
 fi
 
+function test_faucetizer {
+    test_case=$1
+    fc_file=faucet.yaml
+    ns_file=ns_$test_case.yaml
```
grafnu: ns_ is a bad prefix because it's very ambiguous and misleading. Typically it means "namespace" -- the contents of the file descirbe stuff about devices, so I would prefix with "devices_" or something more meaningful.

forch/faucetizer.py
didovesei: This will not be handled by authenticator? Does it mean authenticator will return a device with VLAN but not role?

grafnu:
yes, that is a possible case (role is an optional attribute).

forch/faucetizer.py
didovesei: Added Device class for encapsulation

bin/test_topology
didovesei: changed to state_*

grafnu:
state_ is also too generic.  State of what?  It's an extremely overloaded
word.  It is not, for example, the state of the network.

bin/test_topology
didovesei: Changed to devices

forch/faucetizer.py
```diff
@@ -35,14 +66,12 @@ def process_network_state(self, network_state):
...
-        device_map['new_switch'] = learning.switch
-        device_map['new_port'] = learning.port
+        device = self._devices.setdefault(mac, Device())
+        device.new_learning = learning
 
     def _process_device_behavior(self, mac, behavior):
-        device_map = self._devices.setdefault(mac, {})
-        device_map['vlan'] = behavior.vid
-        device_map['role'] = behavior.role
+        device = self._devices.setdefault(mac, Device)
```
grafnu: Device here and Device() above?

forch/faucetizer.py
```diff
@@ -35,14 +66,12 @@ def process_network_state(self, network_state):
...
+        device.behavior = behavior
```
grafnu: how is this a direct attribute modifier?  I don't see a "behavior" attribute on the class?

forch/faucetizer.py
```diff
@@ -14,6 +14,37 @@
 LOGGER = logging.getLogger('faucetizer')
 
 
+class Device:
```
grafnu: NB: Unless there's significant logic here, you can likely reduce overhead (fewer lines of code) overall just using a PB.  Your call.

forch/faucetizer.py

forch/faucetizer.py
didovesei: Changed to explicit attributes

forch/faucetizer.py
didovesei: There are a few logics (e.g., is_fulfilled(), etc.) encapsulated in the Device class. I think there are more pertinent to a Device itself. So here I would keep Device class instead of using protobuf?

forch/faucetizer.py
```diff
@@ -1,49 +1,182 @@
...
+        device.new_learning = learning
+
+    def _process_device_behavior(self, mac, behavior):
+        device = self._devices.setdefault(mac, Device())
+        device.behavior = behavior
+
+    def process_faucet_config(self, faucet_config):
+        """Process faucet config when base faucet config changes"""
+        self._base_faucet_config = faucet_config
+        self._faucet_config = faucet_config
```
grafnu: Name _faucet_config with a prefix, so like _out_faucet_config or something clear.

forch/faucetizer.py
```diff
@@ -1,49 +1,182 @@
...
+
+
+class Device:
+    """A device with learning and behavior"""
+    def __init__(self):
+        self.old_learning = None
+        self.new_learning = None
+        self.behavior = None
+
+    def has_migrated(self):
```
grafnu: is_dirty

forch/faucetizer.py
```diff
@@ -1,49 +1,182 @@
...
+        """Determine if the switch or port of a device has changed"""
+        if not self.old_learning:
+            return False
+        is_same_switch = self.old_learning.switch != self.new_learning.switch
+        is_same_port = self.old_learning.port != self.new_learning.port
+        return not is_same_switch or not is_same_port
+
+    def is_fulfilled(self):
```
grafnu: is_complete

forch/faucetizer.py
```diff
@@ -1,49 +1,182 @@
...
+        device.behavior = behavior
+
+    def process_faucet_config(self, faucet_config):
+        """Process faucet config when base faucet config changes"""
+        self._base_faucet_config = faucet_config
+        self._faucet_config = faucet_config
+        self._faucetize()
+
+    def _faucetize(self, macs=None):
+        if not self._faucet_config:
```
grafnu: this should be depenedent on the base config existing, not a previous output config, and every time through this function, it should start with a clean (deep) copy of _base_faucet_config

forch/faucetizer.py
```diff
@@ -1,49 +1,182 @@
...
+                self._reset_port_config(device)
+                commit = True
 
-        sorted_macs = list(self._network_state.get('device_mac_behaviors', {}).keys())
-        sorted_macs.sort()
-        sys.stdout.write(f'{sorted_macs}\n')
+            if device.is_fulfilled():
+                switch = device.new_learning.switch
+                port = device.new_learning.port
+                switch_config = self._faucet_config.get('dps', {}).get(switch, {})
```
grafnu: until it's committed, the _faucet_config maniplated in this function should be a local varaible only.  The class variable should reflect the currently written config.

forch/faucetizer.py
```diff
@@ -1,49 +1,182 @@
...
+                                   switch, port)
+                    continue
+                port_config['native_vlan'] = device.behavior.vid
+                role = device.behavior.role
+                if role:
+                    port_config['acls_in'] = [f'role_{role}']
+                commit = True
 
+        if commit:
+            self._commit_faucet_config(macs)
```
grafnu: then pass in new config to be committed here as function argument

forch/faucetizer.py

forch/faucetizer.py

forch/faucetizer.py

forch/faucetizer.py
didovesei: Done. Using a copy of self._out_faucet_config now

forch/faucetizer.py

forch/faucetizer.py
didovesei: I think it should be enough to start with a copy of self._out_faucet_config if we use a local variable in the following processing? This way devices map will be consitent with self._out_faucet_config.

forch/faucetizer.py
```diff
@@ -69,29 +70,30 @@ def _process_device_behavior(self, mac, behavior):
     def process_faucet_config(self, faucet_config):
         """Process faucet config when base faucet config changes"""
         self._base_faucet_config = faucet_config
-        self._faucet_config = faucet_config
+        self._out_faucet_config = faucet_config
```
grafnu: should be None initially, since you don't know the state of the external files.

forch/faucetizer.py
```diff
@@ -69,29 +70,30 @@ def _process_device_behavior(self, mac, behavior):
...
         self._faucetize()
 
     def _faucetize(self, macs=None):
-        if not self._faucet_config:
+        if not self._out_faucet_config:
```
grafnu: _base_faucet_config

forch/faucetizer.py
```diff
@@ -69,29 +70,30 @@ def _process_device_behavior(self, mac, behavior):
...
             return
 
         if not macs:
             macs = self._devices.keys()
 
         commit = False
+        out_faucet_config = self._out_faucet_config
```
grafnu: make copy

forch/faucetizer.py
grafnu: I suppose -- for now it's fine as long as it works. It feels like you're over-optimizing and likely to create corner cases (e.g. state left-over in the out faucet config that isn't accurately reflected in the current setup), which are much harder to deal with than clean logic (even if it's slower).

What happens, for example, if/when the base faucet config changes?

forch/faucetizer.py
```diff
@@ -1,49 +1,193 @@
...
+
+    def process_faucet_config(self, faucet_config):
+        """Process faucet config when base faucet config changes"""
+        self._base_faucet_config = faucet_config
+        self._out_faucet_config = faucet_config
+        self._faucetize()
+
+    def _faucetize(self, macs=None):
+        if not self._out_faucet_config:
+            return
```
grafnu: error not silent return

forch/faucetizer.py
grafnu: Also, I don't see where you "start with a copy of" -- did the copy get missed?

grafnu:
Also, tests are not passing... need to clear that up!

forch/faucetizer.py
```diff
@@ -11,132 +11,72 @@
...
-        if not base_port_config:
-            LOGGER.warning('Switch or port not defined in base faucet config: %s %s', switch, port)
-            return False
-
-        port_config.update(base_port_config)
-
-        return True
-
-    def _commit_faucet_config(self, macs, out_faucet_config):
+    def _generate_faucet_config(self, out_faucet_config):
```
grafnu: To keep the class consistent, this should be a top-level (not class) function write_faucet_config as a peer to load_faucet_config.  Then, this class should just have a method that's something like "get_dynamic_faucet_config" or something, then the caller would do something like write_faucet_config(OUTPUT_PATH, FAUCITIZER.get_dynamic_faucet_config()) -- right now, you're mixing things up because the input is passed as PBs (read externally), but the output is direct to a file -- just want to be consistent in terms of the class API boundary.  The logic shouldn't change -- just instead of writing to a file, it saves it to a class variable (which is then just returned when necessary).

proto/network_state.proto
```diff
@@ -8,6 +8,7 @@ message NetworkState {
 message DeviceLearning {
   string switch = 1;
   int32 port = 2;
+  bool connecting = 3;
```
grafnu: connected not connecting -- state is absolute.

proto/device.proto
```diff
@@ -0,0 +1,11 @@
+/*
```
grafnu: please include in one of the other proto files -- the granularity of "device" is too small.  Rather, things should be grouped by logical bucket for different proto files, and then the "message" scope can be granular.

forch/faucetizer.py

proto/network_state.proto
didovesei: Done

proto/device.proto
didovesei: Moved into network_state.proto

PR #15, opened 2020-01-24
Return error for illegal requests

didovesei:


forch/http_server.py
```diff
@@ -129,7 +129,12 @@ def _split_request(self, base_path, req_path):
         full_path = os.path.join(self._root_path, path)
         if os.path.isdir(full_path):
             full_path = os.path.join(full_path, self._DEFAULT_FILE)
-        return self.read_file(full_path)
+        try:
+            content = self.read_file(full_path)
+            return content
+        except Exception as error:
+            LOGGER.warning("Received illegal request: %s", req_path)
+            return str(error)
```
grafnu: what does this return to the requesting agent (web browser)?  Is it just a string-but-success-code, or a proper 400* or 500* code?

didovesei: Returning appropriate http code now

forch/http_server.py
```diff
@@ -16,6 +16,14 @@
 LOGGER = logging.getLogger('httpserv')
 
 
+class HttpException(Exception):
+    """Http exception base class"""
+    def __init__(self, message, http_status):
+        Exception.__init__(self, message)
+        self.message = message
```
grafnu: self.message is redundant with the Exception base class. Also doesn't seem to be used anywhere?

forch/http_server.py
```diff
@@ -30,32 +38,36 @@ def __init__(self, context, *args, **kwargs):
...
-        self.send_response(200)
-        self.end_headers()
-        parsed = urllib.parse.urlparse(self.path)
-        opts = {}
-        opt_pairs = urllib.parse.parse_qsl(parsed.query)
-        for pair in opt_pairs:
-            opts[pair[0]] = pair[1]
-        message = str(self._context.get_data(self.headers.get('Host'), parsed.path[1:], opts))
-        self.wfile.write(message.encode())
+            LOGGER.warning('Unhandled exception: %s', exception)
```
grafnu: error not warning

grafnu:
Also tests not passing... need to resolve that!

forch/http_server.py
didovesei: Removed

forch/http_server.py
didovesei: Done

didovesei:
> Also tests not passing... need to resolve that!

Updated.

PR #17, opened 2020-01-31
Update Faucetizer to periodically read and write faucet config

didovesei:
…onfig


bin/test_topology
```diff
@@ -5,7 +5,7 @@ source bin/stack_functions
 preamble_setup
 
 # Could be run as its own test shard, if desired.
-bin/test_basics
+#bin/test_basics
```
grafnu: please restore

forch/faucetizer.py
```diff
@@ -91,21 +91,23 @@ def process_devices_state(faucetizer: Faucetizer, devices_state: DevicesState):
 
 def load_faucet_config(file):
     """Load network state file"""
-    LOGGER.info('Loading faucet config file %s', file)
     with open(file) as config_file:
         faucet_config = yaml.safe_load(config_file)
-    LOGGER.info('Loaded faucet config from %s', file)
     return faucet_config
```
grafnu: nb: you can inline the return if you want -- no need for the local variable in this case

forch/forchestrator.py
```diff
@@ -217,6 +245,11 @@ def main_loop(self):
             LOGGER.error("Exception: %s", e)
             raise
 
+    def stop_schedulers(self):
```
grafnu: just stop() -- might eventually stop other things or release other resources

topo/dva/behaviors_block-yellow.yaml
```diff
@@ -0,0 +1,20 @@
+{
+  device_mac_behaviors:
```
grafnu: format as per other files -- either as yaml or with { trailing on previous line

forch/forchestrator.py
```diff
@@ -74,7 +75,7 @@ def initialize(self):
         self._cpn_collector = CPNStateCollector()
 
         self._faucet_config_file = os.path.join(
-            os.getenv('FAUCET_CONFIG_DIR'), _FAUCET_CONFIG_DEFAULT)
+            os.getenv('FAUCET_CONFIG_DIR'), _DYNAMIC_CONFIG_DEFAULT)
```
grafnu: this doesn't seem right -- why is this DYNAMIC config?  How is this entry sync'd with other places the dynamic_config is specified?  Maybe rename varaible to self._structuctural_config_file or self._behavorial_config_file to disambiguate cleanly?  I would make sure that the value is calculated in exactly one place, and cached and used elsewhere (i.e. _DYANMIC_CONFIG_DEFAULT should only be used in one place)

forch/forchestrator.py
```diff
@@ -91,60 +92,87 @@ def initialize(self):
...
+        if not static_behaviors_file:
+            return
+        static_behaviors_path = os.path.join(
+            os.getenv('FAUCET_CONFIG_DIR'), static_behaviors_file)
+        devices_state = faucetizer.load_devices_state(static_behaviors_path)
+        for mac, device_behavior in devices_state.device_mac_behaviors.items():
+            self.process_device_behavior(mac, device_behavior)
+
+    def _initialize_faucetizer(self):
+        if not self._config.get('orchestration', {}).get('run_faucetizer'):
```
grafnu: I would remove run_faucetizer and just key off of dynamic_config_file being defined.  If it's not configed, then there's implicilty no where for the output to go so nothing should happen. Somewhat arbitrary, but just throwing out the idea... in some sense it seems more clean to me (one less configuration variable).

topo/dva/forch.yaml
```diff
@@ -5,8 +5,11 @@ site:
     nz-kiwi-ctl1:
       fqdn: localhost.n
       port: 9020
-static_device_info:
+orchestration:
+    run_faucetizer: True
```
grafnu: this is where I'm thinking the run_faucetizer variable is redundant.... if "dynamic_config_file" is defined, then that's all we need to know!

forch/forchestrator.py
```diff
@@ -91,60 +92,87 @@ def initialize(self):
...
+            self._faucetizer = faucetizer.Faucetizer(structural_config)
+
+        interval = self._config.get('orchestration', {}).get('faucetize_interval', 1)
+        dynamic_config_file = self._config.get('orchestration', {}).get(
+            'dynamic_config_file', _DYNAMIC_CONFIG_DEFAULT)
+        dynamic_config_path = os.path.join(os.getenv('FAUCET_CONFIG_DIR'), dynamic_config_file)
+        self._faucetize_scheduler = HeartbeatScheduler(interval)
+        self._faucetize_scheduler.add_callback(functools.partial(
+            faucetizer.update_structural_config, self._faucetizer, structural_config_path))
+        self._faucetize_scheduler.add_callback(functools.partial(
```
grafnu: does this trigger faucet to try and reload the file each time?  If so, a few things:
1) Make the default period longer. 1s seems overly agressive.
2) File a bug to make this do change detection or based off of inode watching like faucet does.  
This is fine for now, but want to put it in the queue for eventual cleanup.

forch/forchestrator.py
```diff
@@ -91,60 +92,87 @@ def initialize(self):
...
+        structural_config_file = self._config.get('orchestration', {}).get(
+            'structural_config_file', _STRUCTURAL_CONFIG_DEFAULT)
+        structural_config_path = os.path.join(
+            os.getenv('FAUCET_CONFIG_DIR'), structural_config_file)
+        LOGGER.info('Loading structural config from %s', structural_config_path)
+        with open(structural_config_path) as file:
+            structural_config = yaml.safe_load(file)
+            self._faucetizer = faucetizer.Faucetizer(structural_config)
+
+        interval = self._config.get('orchestration', {}).get('faucetize_interval', 1)
```
grafnu: faucetize_interval_sec

forch/forchestrator.py
didovesei: So self._faucet_config_file is the file that is read by faucet. It will be used in _restore_states() and get_sys_config(). So I think here it should be dynamic config file instead of structural config file?

grafnu:
Ok, but then the name should be changed so it's not ambiguous. Also, should
be calculated in only one place since now there's two, and I think they're
using different logic.

forch/forchestrator.py

forch/forchestrator.py

forch/forchestrator.py

forch/forchestrator.py
didovesei: Yes, done

forch/forchestrator.py
didovesei: Changed _DYNAMIC_CONFIG_DEFAULT back to _FAUCET_CONFIG_DEFAULT, and set self._faucet_config_file in _initialize_faucetizer(). Idea is that if forch is configured to run faucetizer, then it will use the dynamic file as self._faucet_config_file. Otherwise, the self._faucet_config_file will be initialized as before.

topo/dva/behaviors_block-yellow.yaml
didovesei: Done

forch/forchestrator.py
didovesei: Changed to 60s

forch/forchestrator.py
didovesei: Renamed self._faucet_config_file to self._behavior_config_file

forch/forchestrator.py
```diff
@@ -97,8 +92,16 @@ def initialize(self):
         self._process_static_device_placement()
         self._process_static_device_behavior()
 
+        if not self._behavior_config_file:
+            self._behavior_config_file = os.path.join(
```
grafnu: you are assigning this variable in two places, it should only be one

forch/forchestrator.py
```diff
@@ -50,7 +50,7 @@ class Forchestrator:
 
     def __init__(self, config):
         self._config = config
-        self._faucet_config_file = None
+        self._behavior_config_file = None
```
grafnu: _behavorial_ not _behavior_

forch/forchestrator.py
```diff
@@ -149,16 +154,14 @@ def _initialize_faucetizer(self):
...
 
-        interval = self._config.get('orchestration', {}).get('faucetize_interval', 1)
-        dynamic_config_file = self._config.get('orchestration', {}).get(
-            'dynamic_config_file', _DYNAMIC_CONFIG_DEFAULT)
-        dynamic_config_path = os.path.join(os.getenv('FAUCET_CONFIG_DIR'), dynamic_config_file)
+        interval = self._config.get('orchestration', {}).get('faucetize_interval_sec', 60)
+        self._behavior_config_file = os.path.join(
+            os.getenv('FAUCET_CONFIG_DIR'), dynamic_config_file)
         self._faucetize_scheduler = HeartbeatScheduler(interval)
         self._faucetize_scheduler.add_callback(functools.partial(
```
grafnu: only one callback. Generally speaking you can't assume the ordering of callbacks, so if you have an atomic operation, it should be one callback.

forch/faucetizer.py
```diff
@@ -91,21 +91,22 @@ def process_devices_state(faucetizer: Faucetizer, devices_state: DevicesState):
...
-    except Exception as error:
-        LOGGER.error('Cannot write faucet config: %s', error)
+def update_structural_config(faucetizer: Faucetizer, file):
+    """Read structural config from file and update in faucetizer"""
+    with open(file) as structural_config_file:
+        structural_config = yaml.safe_load(structural_config_file)
+        faucetizer.process_faucet_config(structural_config)
+
+
+def write_dynamic_config(faucetizer: Faucetizer, file):
```
grafnu: You're calling it _dynamic_ here but _behaviorial_ elsewhere -- likely good to be consistent.

forch/forchestrator.py
```diff
@@ -91,60 +87,95 @@ def initialize(self):
...
+        static_behaviors_file = self._config.get('orchestration', {}).get('static_device_behavior')
+        if not static_behaviors_file:
+            return
+        static_behaviors_path = os.path.join(
+            os.getenv('FAUCET_CONFIG_DIR'), static_behaviors_file)
+        devices_state = faucetizer.load_devices_state(static_behaviors_path)
+        for mac, device_behavior in devices_state.device_mac_behaviors.items():
+            self.process_device_behavior(mac, device_behavior)
+
+    def _initialize_faucetizer(self):
```
grafnu: _attempt_faucitizer_initialize or something similar -- since it may not actually initialize it. (same as authenticator class)

forch/forchestrator.py
didovesei: The callbacks are stored in a list in HeartbeatScheduler, which should gurantee the order. Do you mean the callbacks among different HeartbeatSchedulers?

forch/forchestrator.py
```diff
@@ -38,7 +38,7 @@
 
 _FORCH_CONFIG_DEFAULT = 'forch.yaml'
 _STRUCTURAL_CONFIG_DEFAULT = 'faucet.yaml'
-_DYNAMIC_CONFIG_DEFAULT = 'faucet.yaml'
+_FAUCET_CONFIG_DEFAULT = 'faucet.yaml'
```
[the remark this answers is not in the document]
didovesei: Changed to _BEHAVIORAL_CONFIG_DEFAULT

forch/forchestrator.py
didovesei: Added self._calculate_behavioral_config() where self._behavior_config_file is assigned

forch/forchestrator.py

forch/faucetizer.py
didovesei: Done. Always using behavioral instead of dynamic now.

forch/forchestrator.py
didovesei: Now initializing faucetizer based on the result of self._calculate_behavioral_config().

forch/forchestrator.py
grafnu: You only know that to be true because you know the internal implementation of the other class -- it's breaking the abstraction and bleeding details. If I were to change out the implementation of the scheduler, then that assertion might not be true.  So, yes, it _works_ in this circumstance, but it's _fragile_ and error prone.  Much better to code against the abstraction.   It's trivial to do -- you just create a lambda function and then call the two other functions from there.

didovesei: Done, PTAL

PR #18, opened 2020-02-01
Adding test for minimal-learning default VLAN ACL

didovesei:


bin/test_topology
```diff
@@ -18,17 +18,32 @@ fi
...
 
     $PYTHON -m forch.faucetizer -s $state_file -c $fc_file -o $output_file
     echo Comparing $OUT_DIR/$output_file and $FORCH_CONFIG_DIR/$sample_file
     (diff $OUT_DIR/$output_file $FORCH_CONFIG_DIR/$sample_file && echo 0 || echo 1) >> $TEST_RESULTS
 }
 
+function test_stack {
+    desc=stack-$1
+    echo Starting $desc test... | tee -a $TEST_RESULTS
+    wget localhost:8001 -O- 2>/dev/null | grep 'nz-kiwi-t2sw1' | grep -oE '9a:02:57:1e:8f:01' | tee -a $TEST_RESULTS
```
grafnu: what is this endpoint (8001)?  Is this the faucet varz?

bin/test_topology
```diff
@@ -18,17 +18,32 @@ fi
...
+
 $PYTHON -m forch.topology >> $TEST_RESULTS
 $PYTHON -m forch.authenticator >> $TEST_RESULTS
 
+echo %%% intitial | tee -a $TEST_RESULTS
+test_stack intitial
+
+echo %%% l2-learning | tee -a $TEST_RESULTS
+docker exec forch-faux-1 arp -s 192.168.1.254 02:00:00:00:00:00
+docker exec forch-faux-1 bash -c "echo -n 'abc' > /dev/udp/192.168.1.254/67"
```
grafnu: can you also add a negative test case?  E.g., some request that does NOT work because it's blocked by the ACL?  Likely just need two lines that do something similar for forch-faux-2 but against port 68 (dhcp client port)...  Later on we can test that the request is opened up given an applied role-based ACL, but the doesn't-do-anything test is sufficient for now.

.travis.yml
```diff
@@ -11,14 +11,14 @@ addons:
...
   - set -e
-  - $TEST_SKIP bin/setup_stack local
+  - $SKIP_STACK bin/setup_stack local $MODE
   - bin/run_test_set
 env:
   matrix:
     - TEST_SET=stack
     - TEST_SET=reconfig
-    - TEST_SET=topology TEST_SKIP=true
+    - TEST_SET=topology SKIP_INSTALL=true MODE=dva
```
grafnu: On second thought, I don't think 1) this will work, and 2) it's the right thing to do. First, you have to run the install stuff in order to setup the stack, so you can't do the SKIP_INSTALL part and then try to use the stack. Second, I think we should keep test_topology to be purely functional and not integration. So it can keep the things that are just static file comparisons, but anything that's more dynamic (requires a functioning stack) should go elsewhere.

So, we can either move the work you did here into one of the other test setups, or just create a new one (so there's four instead of three). I think creating a new one will likely be best (although slow)... it can be something like test_access or something (all about testing various dynamic access policies).

bin/test_topology
didovesei: Yes, was going to make it a parameter but forgot..

bin/test_topology
didovesei: OK will add

.travis.yml
didovesei: Created a new test_set

bin/setup_stack
```diff
@@ -22,6 +22,13 @@ else
     shift
 fi
 
+if [ -z "$1" ]; then
```
grafnu: This should be something more specific, not just "is a variable defined" -- e.g. like "local" or "devices" -- also, put before $MODE.

didovesei: So this should be an non-optional argument?

grafnu:
It should be exactly the same as "local" and "devices"

topo/dva/faucet_structural.yaml
```diff
@@ -0,0 +1,248 @@
+dps:
```
grafnu: Why is this faucet_structural.yaml and not just faucet.yaml?  I thought as per the design doc that faucet.yaml, in it's basic form, was the structural one?  Unless there's a clear reason for differentiation now (they're identical files, from what I can tell?) -- better to keep them the same and change them later.

didovesei: My intention was that faucet.yaml is used for minimal VLAN ACL test, while faucet_structural.yaml is used for test_topology as they are different test setups. Since they are the same, I am going to remove faucet_structural.yaml.

didovesei: Removed

PR #19, opened 2020-02-04
Radius query module

anurag6:


forch/radius_query.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __init__(self, listen_ip, listen_port, server_ip,  # pylint: disable=too-many-arguments
+                 server_port):
+        self.socket = None
+        self.listen_ip = listen_ip
+        self.listen_port = listen_port
+        self.server_ip = server_ip
+        self.server_port = server_port
+
+    def setup(self):
```
grafnu: seems like this code is duplicate with radius_socket.py?

forch/radius_query.py
```diff
@@ -0,0 +1,113 @@
...
+    def encode_mab_message(self, src_mac, port_id=None):
+        radius_id = self.get_next_radius_pkt_id()
+        req_authenticator = self.get_req_authenticator()
+        self.packet_id_to_mac[radius_id] = {'src_mac': src_mac, 'port_id': port_id}
+        self.packet_id_to_req_authenticator[radius_id] = req_authenticator
+
+        attr_list = []
+        mac_str = str(src_mac).replace(':', "")
+        attr_list.append(UserName.create(mac_str))
+        attr_list.append(CallingStationId.create(str(src_mac).replace(':', '-')))
```
grafnu: extract to variable like "mac_str" -- and neither one should be called mac_str, qualify each one (mac_username & mac_password?), or inline both.

forch/radius_datatypes.py
```diff
@@ -0,0 +1,263 @@
+"""Radius Attribute Datatypes"""
+import struct
+
+import math
+from forch.utils import MessageParseError
+
+
+class DataType():
+    """Parent datatype class, subclass should provide implementation for abstractmethods.
+    May """
```
grafnu: May?

forch/forchestrator.py
```diff
@@ -34,7 +34,7 @@
 _FAUCET_CONFIG_DEFAULT = 'faucet.yaml'
 _DEFAULT_PORT = 9019
 _PROMETHEUS_HOST = '127.0.0.1'
-_LOG_FORMAT = '%(asctime)s %(name)-8s %(levelname)-8s %(message)s'
+_LOG_FORMAT = '%(asctime)s %(name)-8s %(levelname)-8s %(filename)s:%(lineno)s - %(funcName)20s() %(message)s'
```
grafnu: not convinced this is a good idea -- kinda spammy log messages.  Maybe remove for now?

forch/authenticator.py
```diff
@@ -54,8 +56,16 @@ def process_auth_result(self):
             auth_example = dict_proto(auth_obj, AuthResult)
             sys.stdout.write(str(proto_dict(auth_example)) + '\n')
 
+    def do_mab_request(self):
+        Socket = collections.namedtuple('Socket', 'listen_ip, listen_port, server_ip, server_port')
+        socket_info = Socket('0.0.0.0', 0, '172.17.0.5', 1812)
```
grafnu: as discussed, make command line arguments.  Seems like it should be the server_ip, the 'SECRET' mac address, and nonce.

forch/radius_query.py
anurag6: You are right. Let me correct that.

forch/radius_query.py
anurag6: It is used thrice in the method. Are you sure I should name it different things? Because it is exactly what it is named, MAC addr in str format.

anurag6: I will inline it. It is twice not thrice.

forch/forchestrator.py
anurag6: I agree. That was just a temporary change I made to make debugging easier.

forch/radius_query.py
anurag6: Done.

forch/radius_query.py
anurag6: Done

forch/radius_datatypes.py
anurag6: Done

forch/forchestrator.py
anurag6: Done

PR #20, opened 2020-02-05
Read placements.yaml and do MAB auth for each MAC entry, printing the result in logs

anurag6:
Change currently runs auth for 3 MACs in placements.yaml.


forch/authenticator.py
```diff
@@ -57,14 +66,23 @@ def process_auth_result(self):
...
-        LOGGER.info('sending MAB request')
-        radius_query.send_mab_request(_args.src_mac, _args.port_id)
-        radius_query.receive_radius_messages()
+        LOGGER.info('sending MAB request for %s', src_mac)
+        self.radius_query.send_mab_request(src_mac, port_id)
+
+    def process_device_placement(self, src_mac, device_placement):
+        """Process device placement info and initiate mab query"""
+        if not self.radius_query:
+            LOGGER.warning("RADIUS query module not setup. Ignoring auth request for %s", src_mac)
```
grafnu: this should error out -- this method shouldn't be called if module not setup

forch/forchestrator.py
```diff
@@ -82,6 +86,24 @@ def initialize(self):
         self._cpn_collector.initialize()
         LOGGER.info('Using peer controller %s', self._get_peer_controller_url())
         self._register_handlers()
+
+        r_info = self._config.get('radius_info')
```
grafnu: don't abbreviate

forch/forchestrator.py
```diff
@@ -82,6 +86,24 @@ def initialize(self):
...
+        if r_info:
+            radius_ip = r_info.get('server_ip')
+            radius_port = r_info.get('server_port')
+            secret = r_info.get('secret')
+            if not (radius_ip and radius_port and secret):
+                LOGGER.warning('Invalid radius_info in config.')
```
grafnu: seems like more of an error, also, log the three arguments so somebody might have a clue what's wrong, but use bool(secret) to avoid leakage

forch/forchestrator.py
```diff
@@ -82,6 +86,24 @@ def initialize(self):
...
+            else:
+                self._authenticator = Authenticator(radius_ip, radius_port, secret)
```
grafnu: make sure success is logged, either here or inside of Authenticator. Include radius_ip and raduis_port, but not secret

topo/bond/placement.yaml
```diff
@@ -0,0 +1,23 @@
+{
+    device_mac_placements:
```
grafnu: format with { on end of line

forch/radius_query.py
```diff
@@ -1,8 +1,10 @@
 """Talks and listens to RADIUS. Takes a packet object as input"""
-from queue import Queue
 import logging
 import os
 
+from queue import Queue
```
grafnu: Why is this even here?  Is it used for anything?  If not, please remove.  And if it is... then I'm gonna want to take a deeper look at it :-)

forch/radius_query.py
```diff
@@ -27,6 +29,14 @@ def __init__(self, socket_info, radius_secret):
         self.radius_socket = RadiusSocket(socket_info.listen_ip, socket_info.listen_port,
                                           socket_info.server_ip, socket_info.server_port)
         self.radius_socket.setup()
+        self.radius_socket.lock = RLock()
```
grafnu: what is this for?  a) using .lock is horrific because you're breaking encapsulation, and b) I really don't think it's necessary. How is sending a UDP packet require a lock?

anurag6: Python socket isn't thread safe. The alternative would be having a thread safe queue into which state machines dump their requests into and a worker thread keeps emptying it.

forch/radius_query.py
anurag6: Oh yeah. That's an artifact of the sned queue I mentioned in the comment below. Let me remove it.

forch/authenticator.py
anurag6: I am removing the check altogether, so it errors out

forch/forchestrator.py
anurag6: done.

forch/forchestrator.py
anurag6: done.

forch/authenticator.py
anurag6: done.

topo/bond/placement.yaml
anurag6: done

forch/radius_query.py
anurag6: done

forch/forchestrator.py
anurag6: done.

forch/radius_query.py
anurag6: MAde changes to isolate lock in socket. Is it ok?

grafnu: yes, looks about right!

topo/bond/placement.yaml
```diff
@@ -0,0 +1,19 @@
+{
+    device_mac_placements:{
```
grafnu: space between : and {

topo/dva/placement.yaml
```diff
@@ -0,0 +1,23 @@
+{
+    device_mac_placements:
```
grafnu: format as per other file

forch/forchestrator.py
```diff
@@ -85,7 +85,18 @@ def initialize(self):
...
+        device_info = self._config.get('static_device_info', {})
+        if 'static_device_placement' in device_info:
+            placement_file = os.path.join(
+                os.getenv('FAUCET_CONFIG_DIR'), device_info['static_device_placement'])
+            device_placement_info = yaml_proto(placement_file, DevicesState).device_mac_placements
+            for eth_src, device_placement in device_placement_info.items():
+                self.process_device_placement(eth_src, device_placement)
+        self._initialized = True
+
+    def _create_authenticator(self):
```
grafnu: this doesn't actually create the authenticator, it *maybe* creates authenticator. So, either
1) call it _maybe_create_authenticator()
2) only call _create_authenticator() if 'radius_info' is defined (and then pass in radius_info itself)

PR #21, opened 2020-02-05
Load device behavior input

didovesei:


forch/forchestrator.py
```diff
@@ -81,13 +86,23 @@ def initialize(self):
         self._local_collector.initialize()
         self._cpn_collector.initialize()
         LOGGER.info('Using peer controller %s', self._get_peer_controller_url())
+
+        self._faucetizer = Faucetizer()
```
grafnu: this should be conditional somehow... e.g. based on something in the forch.yaml file, and if not enabled then _faucetizer should be None

forch/forchestrator.py
```diff
@@ -87,10 +88,20 @@ def initialize(self):
...
+        structural_config_file = self.config.get('dva', {}).get(
+            'structural_config_file', _STRUCTURAL_CONFIG_DEFAULT)
+        structural_config_path = os.path.join(
+            os.getenv('FAUCET_CONFIG_DIR'), structural_config_file)
+        if self._config.get('dva', {}).get('run_faucetizer', False):
+            with open(structural_config_path) as file:
+                structural_config = yaml.safe_load(file)
+                self._faucetizer = faucetizer.Faucetizer(structural_config)
+
+        static_behaviors_file = self._config.get('dva', {}).get('static_device_behavior')
```
grafnu: dva is too specific since all this stuff, for example, also applies to roles and other device behavior considerations. Maybe put them under an "orchestration" bucket.

also, above you have the config dir prefixed by FAUCET_CONFIG_DIR env but not here?  Seems like both should have a similar treatment (even if with different env variables)

Also add logging that says "loading structural config from %s" -- so it's clear to people which actual file is being used.

forch/forchestrator.py

PR #24, opened 2020-02-10
Enable Faucetizer to receive and process segment to vlan mappings

didovesei:


forch/faucetizer.py
```diff
@@ -40,6 +40,10 @@ def process_device_behavior(self, eth_src, behavior):
             device = self._devices.setdefault(eth_src, Device())
             device.behavior.CopyFrom(behavior)
 
+    def process_segments_to_vlans(self, segments_to_vlans):
+        """Process single segment vlan mapping"""
```
grafnu: if it's single segment, then it should be segment_to_vlan (not plurals). Where is this function used? I don't see it anywhere in the PR so I can't tell what the intended use is.

forch/faucetizer.py
```diff
@@ -61,7 +65,11 @@ def _faucetize(self):
                                    device.placement.switch, device.placement.port)
                     continue
 
-                port_cfg['native_vlan'] = device.behavior.vid
+                vid = self._segments_to_vlans.get(device.behavior.segment)
+                if not vid:
+                    raise Exception('Device segment does not have a matching vlan %s %s',
```
grafnu: this will bork the entire configuration, which is likely overkill.  I would log a warning (since it's not really an error in the system, but one of maybe how RADIUS is configured), and then punt the device to the initialization vlan.

forch/forchestrator.py
```diff
@@ -158,7 +159,14 @@ def _initialize_faucetizer(self):
         LOGGER.info('Loading structural config from %s', structural_config_path)
         with open(structural_config_path) as file:
             structural_config = yaml.safe_load(file)
-            self._faucetizer = faucetizer.Faucetizer(structural_config)
+
+        segments_vlans_file = self._config.get('orchestration', {}).get(
+            'segments_vlans_file', _SEGMENTS_VLAN_DEFAULT)
+        segments_vlans_path = os.path.join(os.getenv('FAUCET_CONFIG_DIR'), segments_vlans_file)
+        LOGGER.info('Loading segment to vlan mappings')
```
grafnu: include actual path in logging statement

forch/faucetizer.py
grafnu: I would also exctact this to a function like self._get_segment_for_behavior(....) and encapsulate everything there

grafnu: Sorry, should be _get_vlan_for_behavior, so it's just:

vid = self._get_vlan_for_behavior(device.behavior)

PR #25, opened 2020-02-11
Connect learning events to process_device_placement

anurag6:


forch/faucet_state_collector.py
```diff
@@ -1016,3 +1022,7 @@ def _get_egress_port(self, switch):
             if port_attr.get('type') == 'egress':
                 return port
         return None
+
+    def add_placement_callback(self, callback):
```
grafnu: As written, this should be set_placement_callback -- add_ would be a list of callbacks

anurag6: Done

PR #27, opened 2020-02-11
Handle L2_EXPIRE events

anurag6:


anurag6:
@grafnu PTAL. Tests in progress.

forch/faucet_state_collector.py
```diff
@@ -823,6 +823,23 @@ def process_port_learn(self, timestamp, name, port, mac, ip_addr):
                 devices_placement = DevicePlacement(switch=name, port=port, connected=True)
                 self._placement_callback(mac, devices_placement)
 
+    @_dump_states
+    # pylint: disable=too-many-arguments
```
grafnu: do you know why this is required but the one for the method below does not?   They both have 5 arguments, right?

forch/faucet_state_collector.py
```diff
@@ -823,6 +823,23 @@ def process_port_learn(self, timestamp, name, port, mac, ip_addr):
...
+    def process_port_expire(self, timestamp, name, port, mac):
+        """process port learn event"""
```
grafnu: update comment

forch/faucet_state_collector.py
```diff
@@ -823,6 +823,23 @@ def process_port_learn(self, timestamp, name, port, mac, ip_addr):
...
+        with self.lock:
+            LOGGER.info('Learned entry %s at %s:%s expired.', mac, name, port)
+            if self.learned_macs:
```
grafnu: when is learned_macs ever not defined?

forch/faucet_state_collector.py
```diff
@@ -823,6 +823,23 @@ def process_port_learn(self, timestamp, name, port, mac, ip_addr):
...
+                entry = self.learned_macs.pop(mac, None)
```
grafnu: if not self.learned_macs.pop():

forch/faucet_state_collector.py
```diff
@@ -823,6 +823,23 @@ def process_port_learn(self, timestamp, name, port, mac, ip_addr):
...
+                if not entry:
+                    LOGGER.debug('Entry %s doesnt exist in learned macs dict', mac)
+            if name in self.switch_states:
+                learned_macs = self.switch_states[name][LEARNED_MACS]
+                if mac in learned_macs:
+                    learned_macs.remove(mac)
+                else:
+                    LOGGER.debug('Entry %s doesnt exist in learned macs set', mac)
```
grafnu: include name in message
should be warning?  When might this happen?

forch/faucet_state_collector.py
```diff
@@ -823,6 +823,23 @@ def process_port_learn(self, timestamp, name, port, mac, ip_addr):
...
+    @_dump_states
+    # pylint: disable=too-many-arguments
+    def process_port_expire(self, timestamp, name, port, mac):
+        """process port learn event"""
+        with self.lock:
+            LOGGER.info('Learned entry %s at %s:%s expired.', mac, name, port)
+            if self.learned_macs:
+                entry = self.learned_macs.pop(mac, None)
+                if not entry:
+                    LOGGER.debug('Entry %s doesnt exist in learned macs dict', mac)
```
grafnu: warning?

forch/faucet_state_collector.py
```diff
@@ -823,6 +823,23 @@ def process_port_learn(self, timestamp, name, port, mac, ip_addr):
...
+            if name in self.switch_states:
```
grafnu: when would name never not be in switch_states?

forch/faucet_state_collector.py
anurag6: We get a learning expire event for each point at which it was learned. I used to debug behavior so made it debug. Need not be a warning. Can remove completely if needed.

forch/faucet_state_collector.py
anurag6: same as above.

forch/faucet_state_collector.py
grafnu: You didn't answer the question -- when does this case (man not in learned_macs) happen?  Seems like it shouldn't happen, which is why I'm suggesting a warning -- else it's pointing to some other underlying problem in the algorithm?

What about the case where it expires on one switch but not the others?

forch/faucet_state_collector.py
anurag6: cleaned up.

forch/faucet_state_collector.py
anurag6: It happens if the mac is learnt at multiple switches. So you get multiple L2_EXPIRE events. The first one deletes it. The second one would not find the entry. I dont think we should be raising a warning for that.

forch/faucet_state_collector.py
anurag6: cleaned up.

forch/faucet_state_collector.py
anurag6: cleaned up.

forch/faucet_state_collector.py
anurag6: cleaned up.

forch/faucet_state_collector.py
anurag6: cleaned up.

forch/faucet_state_collector.py
grafnu: What about the case where it expires on one switch but not the others?

anurag6: yup. you are right. Fixed code to remove entry from set only when the last entry is removed.

forch/faucet_state_collector.py
```diff
@@ -823,6 +823,23 @@ def process_port_learn(self, timestamp, name, port, mac, ip_addr):
...
+    def process_port_expire(self, timestamp, name, port, mac):
+        """process port expire event"""
+        with self.lock:
+            LOGGER.info('Learned entry %s at %s:%s expired.', mac, name, port)
+            learned_macs = self.switch_states[name][LEARNED_MACS]
+            if mac in learned_macs:
+                learned_macs.remove(mac)
+            else:
+                LOGGER.warning('Entry %s doesnt exist in learned macs set', mac)
+            is_last_mac_entry = True
```
grafnu: Much cleaner to be direct about it, rather than indirectly calculating when to remove it -- learned_macs contains a map of switches, so each time through you remove the switch that expired for this mac, and then if that map is empty you can delete the entire entry.

anurag6: yup. Done.

PR #28, opened 2020-02-12
Forch initialization enhancements

didovesei:
* Generate behavioral faucet config file during forch initialization
* Wait for faucet to load config to finish intitialization, to avoid hash comparison assert

### DIFF
--- a/forch/forchestrator.py
+++ b/forch/forchestrator.py
@@ -224,8 +224,7 @@ class Forchestrator:
                 event.timestamp, event.dp_name, event.port_no, event.eth_src, event.l3_src_ip)),
         ])
 
-    def _get_varz_config(self):
-        metrics = self._varz_collector.get_metrics()
+    def _get_varz_config(self, metrics):
         varz_hash_info = metrics['faucet_config_hash_info']
         assert len(varz_hash_info.samples) == 1, 'exactly one config hash info not found'
         varz_config_hashes = varz_hash_info.samples[0].labels['hashes']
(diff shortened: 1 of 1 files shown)


forch/forchestrator.py
```diff
@@ -214,17 +224,25 @@ def _register_handlers(self):
...
         # loss of events inbetween.
         assert self._faucet_events.event_socket_connected, 'restore states without connection'
-        metrics = self._varz_collector.get_metrics()
 
         # Restore config first before restoring all state from varz.
-        varz_hash_info = metrics['faucet_config_hash_info']
-        assert len(varz_hash_info.samples) == 1, 'exactly one config hash info not found'
-        varz_config_hashes = varz_hash_info.samples[0].labels['hashes']
+        metrics, varz_config_hashes, varz_config_error = self._get_varz_config()
+        if varz_config_error:
```
grafnu: just make the get_varz_config function raise the exception itself, then if you need to retry, you can catch the exceptopn and try again.  I think that's cleaner since there's other reasons _get_varz_config might throw an exceptio

forch/forchestrator.py
```diff
@@ -94,6 +94,15 @@ def initialize(self):
         self._attempt_authenticator_initialise()
         self._process_static_device_placement()
         self._process_static_device_behavior()
+        if self._faucetizer:
+            faucetizer.write_behavioral_config(self._faucetizer, self._behavioral_config_file)
+
+        # wait for faucet to load config
+        while True:
+            time.sleep(1)
```
grafnu: increase timeout (10s?) and add logging message, so it's obvious if the system is stuck on this for whatever reason.

forch/forchestrator.py

forch/forchestrator.py
didovesei: Changed to 10s and added error logging

forch/forchestrator.py
```diff
@@ -214,17 +225,26 @@ def _register_handlers(self):
...
+        metrics = self._varz_collector.get_metrics()
+        varz_hash_info = metrics['faucet_config_hash_info']
+        assert len(varz_hash_info.samples) == 1, 'exactly one config hash info not found'
+        varz_config_hashes = varz_hash_info.samples[0].labels['hashes']
+        varz_config_error = varz_hash_info.samples[0].labels['error']
+
+        if varz_config_error:
+            raise Exception(f'Varz config error: {varz_config_error}')
+
+        return metrics, varz_config_hashes, varz_config_error
```
grafnu: no need to return varz_config_error -- it will always be None

PR #31, opened 2020-02-13
Add session and state machine functionality

anurag6:


forch/faucet_state_collector.py
```diff
@@ -839,6 +839,10 @@ def process_port_expire(self, timestamp, name, port, mac):
                 self.learned_macs[mac][MAC_LEARNING_SWITCH].pop(name)
                 if not self.learned_macs[mac][MAC_LEARNING_SWITCH]:
                     self.learned_macs.pop(mac)
+            port_attr = self._get_port_attributes(name, port)
+            if port_attr and port_attr['type'] == 'access' and self._placement_callback:
```
grafnu: this same logic is now duplicated in two places... extract to common method

forch/mab_state_machine.py
```diff
@@ -0,0 +1,159 @@
+"""Module that implements MAB state machine"""
```
grafnu: state machine is not specific to MAB... it's the device authentication status. MAB is just one means of authentication, but everything else applies to other techniques.

forch/mab_state_machine.py
```diff
@@ -0,0 +1,159 @@
...
+
+import logging
+from threading import Timer
+
+LOGGER = logging.getLogger('mabsm')
+
+
+class State():
```
grafnu: way too complex. I don't understand why all this is necessary for 4 states that have only a handful of transitions. This level of abstraction really isn't needed.  Please see the faucet of daq state machines for simple ways of handling this.  It really can/should just be one simple variable that says "this is my current state". When you want to change it based off of some event, you simply check the current state, and then set the new state appropriate, or log a warning.  Maybe create an intermediate helper method _state_transition() -- ala DAQ.

forch/mab_state_machine.py
```diff
@@ -0,0 +1,159 @@
...
+        self.current_state = self.states[self.UNAUTH]
+
+    def _radius_timeout(self):
+        self.process_trigger(self.REQ_TIMEOUT)
+
+    def _auth_session_timeout(self):
+        self.process_trigger(self.AUTH_TIMEOUT)
+
+    def _handle_accept(self):
+        self.session.session_result(accept=True)
```
grafnu: another indication that the other (session) class isn't helping -- here you just call straight into it. It would be clearner just to be able to say:  self.session_result(accept=True) without the intermediate

forch/mab_state_machine.py
anurag6: But the state machine is MAB. The states and transitions are specific.

grafnu: The only aspect of it that's MAB specific is the radius request and associated trigger. This entire structure would have value, for example, if we wanted to do something based off of IP address. The core of the state machine is managing the radius authenticated lifecycle, right?  The way it's named, if there was another system that needed to do IP-based authentication, or EAPoL-based authentication it would be another class entirely, which I don't think should be necessary.

None of the actual states involved are specific to MAB:

    SENT_REQ = "Sent MAB request"
    RECV_ACCPT = "Received RADIUS accept"
    RECV_REJ = "Received RADIUS reject"
    REQ_TIMEOUT = "RADIUS request timeout"
    AUTH_TIMEOUT = "Auth session timed out"
    EXPIRE = "Host expired"

PR #32, opened 2020-02-13
Initiating integration test for DVA

didovesei:


bin/test_access
```diff
@@ -4,22 +4,48 @@ source bin/stack_functions
...
+echo %%% minimal-learning | tee -a $TEST_RESULTS
 docker exec forch-faux-1 arp -s 192.168.1.254 02:00:00:00:00:00
 docker exec forch-faux-2 arp -s 192.168.1.254 02:00:00:00:00:00
 docker exec forch-faux-1 bash -c "echo -n 'abc' > /dev/udp/192.168.1.254/67"
 docker exec forch-faux-2 bash -c "echo -n 'abc' > /dev/udp/192.168.1.254/68"
 sleep 1
-test_access l2-learning
+test_access minimal-learning
+
+restart_forch
```
grafnu: why is this necessary?

bin/run_faucet
```diff
@@ -32,6 +33,10 @@ if [ "$1" == "gauge" ]; then
     shift
 fi
 
+if [ "$1" == "dva" ]; then
+    FAUCET_CONFIG=faucet_dynamic.yaml
+    shift
+
```
grafnu: missing fi

bin/test_access
didovesei: So basically I wanted to test if forch can generate the behavioral config correctly and the traffic behaves correctly after faucet reloading new config, just to simulate the real case.

grafnu: But why restart it?   Shouldn't it be dynamically applying updates as the state of the system changes?  I would have exected forch to be started at the beginning of the sequence, and then the tests would be monitoring for the changes as they are dynamically applies.  This is, e.g., how the stack tests work with forch -- there it's just testing around links going up and down... ?

didovesei: > why restart it?
Actually that line starts forch for the first time, although it is called "restart_forch"

> I would have exected forch to be started at the beginning of the sequence, and then the tests would be monitoring for the changes as they are dynamically applies.
I Agree. The problem was that in this test setup forch is configured to load placements and behaviors upon starting. If starting forch at the very beginning of the test, the minimal-learning-acl will be overwritten, and the miminal-learning-vlan would not be tested. If instead we do not load placements and behaviors, it is not possible to load it later.

I guess we can either 1) monitoring placements.yaml and behaviors.yaml changes i Forch code and auto reload them, or 2) modify behaviors.yaml such that faux-1 and faux-2 stay on the default VLAN while faux-3 is assigned with a new VLAN so that we can test both minimal-learning and forch behavior?

PR #37, opened 2020-02-14
Update to handle unknown lacp state, preparing for future faucet

grafnu:


forch/faucet_state_collector.py
```diff
@@ -780,14 +780,15 @@ def _get_egress_state_detail(self, links):
...
-            if status.get(LINK_STATE) == LacpState.LacpState.Name(LacpState.active):
+            link_state = status.get(LINK_STATE)
+            links_status.add(LacpState.LacpState.Value(link_state))
+            if link_state == LacpState.LacpState.Name(LacpState.active):
                 egress_name = key
-            if status.get(LINK_STATE) == LacpState.LacpState.Name(LacpState.down):
+            elif link_state == LacpState.LacpState.Name(LacpState.down):
                 link_down = key
-        if links_status == set([LacpState.active, LacpState.up]):
+        if links_status == set([LacpState.active, LacpState.up, LacpState.unknown]):
```
anurag6: Links status set is basically the status of egress links, right? Thy would we have a third link?

grafnu: It's doing a set comparison. Basically it collects a list of all the "allowed" state, and as long as they match the set it goes.  So, it's basically saying that ALL links need to be one of { active, up, unknown } -- notice that down or broken or whatever isn't in there.

anurag6: Current implementation assumes that this event gets called only for egress links but that is not true anymore.

PR #40, opened 2020-02-18
Add timeouts to session states

anurag6:


forch/authenticator.py
```diff
@@ -33,7 +34,9 @@ def __init__(self, radius_ip, radius_port, radius_secret, auth_callback=None):
             self.radius_query = r_query.RadiusQuery(
                 socket_info, radius_secret, self.process_radius_result)
             threading.Thread(target=self.radius_query.receive_radius_messages, daemon=True).start()
-
+        self.timer = threading.Timer(STATE_MACHINE_TIMER_FREQ, self.handle_sm_timeout)
```
grafnu: you should use the heartbeat class that Yufeng created

forch/simple_auth_state_machine.py
```diff
@@ -1,39 +1,95 @@
...
+            message = 'state was %s expected %s' % (self.current_state, expected)
+            assert self.current_state == expected, message
+        LOGGER.info('Transition: %s -> %s', self.current_state, target)
+        self.current_state = target
+
+    def _reset_state_machine(self):
+        self._state_transition(self.UNAUTH)
+        self.retries = 0
+        self.current_timeout = time.time() + self.REJECT_STATE_TIMEOUT
+        self.auth_callback(self.src_mac)
```
grafnu: this needs more than default arguments to indicate failure -- I would suggest passing None/None or whatever to clearly indicate what is going on. Currently, it looks like it's defaulting to some kind of success situation.

forch/simple_auth_state_machine.py
```diff
@@ -1,39 +1,95 @@
...
+            if self.current_state != self.UNAUTH:
+                self._reset_state_machine()
+            self._state_transition(self.REQUEST, self.UNAUTH)
+            self.radius_query_callback(self.src_mac, self.port_id)
 
     def host_expired(self):
         """Host expired"""
-        self.auth_callback(self.src_mac)
+        with self.transition_lock:
+            self.auth_callback(self.src_mac)
```
grafnu: for consistency you should set the state appropriately here too.  You're assuming that the caller is doing the right thing and deleting the object after this call, but it's general good practice to assume that somebody might screw it up.

forch/simple_auth_state_machine.py
```diff
@@ -1,39 +1,95 @@
...
 
 class AuthStateMachine():
     """Class represents the state machine that handles the Auth for a session"""""
 
+    UNAUTH = "Unauthorized"
+    REQUEST = "RADIUS Request"
+    ACCEPT = "Authorized"
+    MAX_RADIUS_RETRIES = 5
+    RADIUS_QUERY_TIMEOUT = 10
+    REJECT_STATE_TIMEOUT = 20
```
grafnu: very inconsistant names -- one is _QUERY_ one is _STATE_ and one is _SESSION_... try to provide a coherent naming structure.

forch/simple_auth_state_machine.py
```diff
@@ -1,39 +1,95 @@
...
         self.radius_query_callback = radius_query_callback
         self.current_state = None
+        self.retries = 0
+        self.current_timeout = 0
+        self.transition_lock = Lock()
+        self._reset_state_machine()
+
+    def _increment_retries(self):
+        if self.retries < self.MAX_RADIUS_RETRIES:
+            self.retries += 1
```
grafnu: it's not really the number of retries, since it's capped -- maybe ._retry_backoff?

forch/simple_auth_state_machine.py
anurag6: Addressed.

forch/simple_auth_state_machine.py
anurag6: addressed.

forch/authenticator.py
```diff
@@ -73,6 +75,11 @@ def process_auth_result(self):
             auth_example = dict_proto(auth_obj, AuthResult)
             sys.stdout.write(str(proto_dict(auth_example)) + '\n')
 
+    def stop_timer(self):
```
grafnu: just stop(self) -- might eventually also stop other things

forch/simple_auth_state_machine.py
grafnu: I meant the variable name (and should be private)

PR #43, opened 2020-02-20
Ignore dynamic auth result for devices in static file

didovesei:


forch/faucetizer.py
```diff
@@ -19,29 +19,44 @@
...
-        self._devices = {}
+        self._dynamic_devices = {}
+        self._persisted_devices = {}
         self._segments_to_vlans = segments_to_vlans
         self._structural_faucet_config = structural_faucet_config
         self._behavioral_faucet_config = None
         self._lock = threading.Lock()
 
-    def process_device_placement(self, eth_src, placement):
+    def process_device_placement(self, eth_src, placement, persisted):
```
grafnu: I would default this to False

forch/faucetizer.py
```diff
@@ -19,29 +19,44 @@
...
             else:
-                self._devices.pop(eth_src, None)
-
-    def process_device_behavior(self, eth_src, behavior):
-        """Process device placement"""
+                removed = self._devices.pop(eth_src, None)
+                if removed:
+                    LOGGER.info('Removed %s placement: %s', device_type, eth_src)
+
+    def process_device_behavior(self, eth_src, behavior, persisted):
```
grafnu: default

forch/forchestrator.py
```diff
@@ -194,17 +194,17 @@ def initialized(self):
         """If forch is initialized or not"""
         return self._initialized
 
-    def _process_device_placement(self, eth_src, device_placement):
+    def _process_device_placement(self, eth_src, device_placement, persisted=False):
```
grafnu: use default  (also, should be consistent -- either used named arg or positional)

didovesei: This is function definition?

forch/faucetizer.py

forch/faucetizer.py

forch/forchestrator.py
```diff
@@ -137,7 +137,7 @@ def _process_static_device_placement(self):
             os.getenv('FAUCET_CONFIG_DIR'), static_placement_file)
         device_placement_info = yaml_proto(placement_file, DevicesState).device_mac_placements
         for eth_src, device_placement in device_placement_info.items():
-            self._process_device_placement(eth_src, device_placement)
+            self._process_device_placement(eth_src, device_placement, True)
```
grafnu: used named argument -- a floating "True" isn't very readable... so static=True

forch/forchestrator.py
```diff
@@ -194,17 +194,17 @@ def initialized(self):
...
+            self._faucetizer.process_device_placement(eth_src, device_placement, static)
         if self._authenticator:
             self._authenticator.process_device_placement(eth_src, device_placement)
 
-    def _process_device_behavior(self, mac, device_behavior):
+    def _process_device_behavior(self, mac, device_behavior, static=False):
         """Function interface of processing device behavior"""
         if self._faucetizer:
-            self._faucetizer.process_device_behavior(mac, device_behavior)
+            self._faucetizer.process_device_behavior(mac, device_behavior, static)
```
grafnu: static=static

forch/forchestrator.py

forch/forchestrator.py

PR #45, opened 2020-02-24
Expose DVA state in NOAH

didovesei:


forch/dva_state_collector.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def write_dva_state(dva_state, file):
+    """Write dva state to file"""
+    dva_state_json = proto_dict(dva_state)
+    with open(file, 'w') as dva_state_file:
+        json.dump(dva_state_json, dva_state_file)
+
+
+def parse_args(raw_args):
```
grafnu: I don't think this is a good approach -- not every module should have args parser -- it should only be for a few seslect entry points. This feature/modules isn't big enough to merit its own args.

forch/dva_state_collector.py
```diff
@@ -0,0 +1,94 @@
...
+    parser.add_argument('-c', '--behavioral-config', type=str, default='faucet.yaml',
+                        help='behavioral faucet config input')
+    parser.add_argument('-o', '--output', type=str, default='faucet.yaml',
+                        help='dva state')
+    return parser.parse_args(raw_args)
+
+
+if __name__ == '__main__':
+    configure_logging()
+    ARGS = parse_args(sys.argv[1:])
```
grafnu: This should be much simpler -- really it's just "test this feature" -- shouldn't be a completely configurable stand alone module -- the code overhead is too high and not worth it.  (It's find for this *one* instance... but imagine doing this 10x ??)

forch/forchestrator.py
```diff
@@ -625,6 +641,7 @@ def main():
         http_server.map_request('host_path', forchestrator.get_host_path)
         http_server.map_request('list_hosts', forchestrator.get_list_hosts)
         http_server.map_request('sys_config', forchestrator.get_sys_config)
+        http_server.map_request('dva_state', forchestrator.get_dva_state)
```
grafnu: THis should not be a separate top-level API... it's just another attribute for a given port and/or MAC address. We already have plenty of entry points to get that information, so it should be an incremental addition to those places, not something new.

For every switch port (already existing API), we should be able to see which VLAN & ACL are applied.

For each MAC address (already existing API), we should be able to see what switch/port/VLAN/ACL it's associated with.

didovesei: Moved VLAN/ACLs into switch_state and list_hosts API.

PR #46, opened 2020-02-24
Move radius_info under auth_config

anurag6:
First draft which moves radius_info field under orchestration. Will add more fields for other constants.

### DIFF
--- a/forch/authenticator.py
+++ b/forch/authenticator.py
@@ -19,7 +19,7 @@ from forch.proto.authentication_pb2 import AuthResult
 LOGGER = logging.getLogger('auth')
 AUTH_FILE_NAME = 'auth.yaml'
 
-STATE_MACHINE_TIMER_FREQ = 3
+STATE_MACHINE_TIMER_INTERVAL = 3
 
 class Authenticator:
     """Authenticate devices using MAB/dot1x"""
@@ -29,8 +29,6 @@ class Authenticator:
         self.sessions = {}
         self.auth_callback = auth_callback
         radius_info = auth_config.get('radius_info')
-        if not radius_info:
-            return
         radius_ip = radius_info.get('server_ip')
         radius_port = radius_info.get('server_port')
         secret = radius_info.get('secret')
@@ -46,7 +44,8 @@ class Authenticator:
             socket_info, secret, self.process_radius_result)
         threading.Thread(target=self.radius_query.receive_radius_messages, daemon=True).start()
 
-        self.timer = HeartbeatScheduler(STATE_MACHINE_TIMER_FREQ)
+        interval = auth_config.get('sm_timer_interval', STATE_MACHINE_TIMER_INTERVAL)
+        self.timer = HeartbeatScheduler(interval)
         self.timer.add_callback(self.handle_sm_timeout)
         self.timer.start()
         LOGGER.info('Created Authenticator module with radius IP %s and port %s.',
@@ -155,7 +154,7 @@ def parse_args(raw_args):
 if __name__ == '__main__':
     configure_logging()
     ARGS = parse_args(sys.argv[1:])
-    auth_config = {
+    AUTH_CONFIG = {
         'auth_config': {
             'radius_info': {
                 'server_ip': ARGS.server_ip,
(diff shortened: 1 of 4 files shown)


forch/authenticator.py
```diff
@@ -23,22 +23,34 @@
...
         self.auth_callback = auth_callback
-        if radius_ip and radius_port and radius_secret:
-            Socket = collections.namedtuple(
-                'Socket', 'listen_ip, listen_port, server_ip, server_port')
-            socket_info = Socket('0.0.0.0', 0, radius_ip, radius_port)
-            self.radius_query = r_query.RadiusQuery(
-                socket_info, radius_secret, self.process_radius_result)
-            threading.Thread(target=self.radius_query.receive_radius_messages, daemon=True).start()
+        radius_info = auth_config.get('radius_info')
+        if not radius_info:
```
grafnu: what does it mean for authentication to exist when this is not defined?   Usually better to fail hard, e.g. not allow the creation of an unconfigured authenticator object, unless there's a structural reason for it.

forch/forchestrator.py
```diff
@@ -114,20 +114,8 @@ def initialize(self):
...
-        if not (radius_ip and radius_port and secret):
-            LOGGER.warning('Invalid radius_info in config. \
-                           Radius IP: %s; Radius port: %s Secret present: %s',
-                           radius_ip, radius_port, bool(secret))
-            raise ConfigError
-        self._authenticator = Authenticator(radius_ip, radius_port, secret, self.handle_auth_result)
-        LOGGER.info('Created Authenticator module with radius IP %s and port %s.',
-                    radius_ip, radius_port)
+        auth_config = self._config.get('orchestration', {}).get('auth_config')
+        self._authenticator = Authenticator(auth_config, self.handle_auth_result)
```
grafnu: this is no longer "attempt" because it does it all the time -- however, as per my other comment I'd leave the function name as is, but only call Authenticator() if auth_config is defined.

forch/authenticator.py
anurag6: Done.

forch/simple_auth_state_machine.py
```diff
@@ -13,24 +13,29 @@ class AuthStateMachine():
...
-    def __init__(self, src_mac, port_id, radius_query_callback, auth_callback):
+    # pylint: disable=too-many-arguments
+    def __init__(self, src_mac, port_id, auth_config, radius_query_callback, auth_callback):
         self.src_mac = src_mac
         self.port_id = port_id
         self.auth_callback = auth_callback
         self.radius_query_callback = radius_query_callback
         self.current_state = None
         self._retry_backoff = 0
         self.current_timeout = 0
```
didovesei: Maybe this can be a private member (name starting with "_")?

forch/simple_auth_state_machine.py
```diff
@@ -13,83 +13,89 @@ class AuthStateMachine():
...
-        self.current_state = None
+        self._auth_callback = auth_callback
+        self._radius_query_callback = radius_query_callback
+        self._current_state = None
         self._retry_backoff = 0
-        self.current_timeout = 0
-        self.transition_lock = Lock()
+        self._current_timeout = 0
+        self._max_radius_backoff = auth_config.get('max_radius_backoff', self.MAX_RADIUS_BACKOFF)
+        self._query_state_timeout = auth_config.get('query_state_timeout', self.QUERY_STATE_TIMEOUT)
```
grafnu: configs should be suffixed with units (_ms? _sec?) -- also drop "state" -- e.g., "query_timeout_ms"

forch/simple_auth_state_machine.py
anurag6: done

forch/simple_auth_state_machine.py
anurag6: Does this need to be changed in the config fields as well or only the variables?

grafnu:
I only care about the external config file... Variables internal to the
class are less important, but I was ask why they were different.

grafnu:
Only care --> mostly care

forch/simple_auth_state_machine.py
anurag6: Done

forch/authenticator.py
```diff
@@ -12,33 +12,45 @@
...
+                           radius_ip, radius_port, bool(secret))
+            raise ConfigError
+        Socket = collections.namedtuple(
+            'Socket', 'listen_ip, listen_port, server_ip, server_port')
+        socket_info = Socket('0.0.0.0', 0, radius_ip, radius_port)
+        self.radius_query = r_query.RadiusQuery(
+            socket_info, secret, self.process_radius_result)
+        threading.Thread(target=self.radius_query.receive_radius_messages, daemon=True).start()
+
+        interval = auth_config.get('sm_timer_interval', STATE_MACHINE_TIMER_INTERVAL)
```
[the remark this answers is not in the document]
anurag6: Done

anurag6:
@grafnu  PTAL

PR #47, opened 2020-02-26
Integration test for auth module

anurag6:


etc/freeradius/certs/index.txt.attr.old
```diff
@@ -0,0 +1 @@
+unique_subject = yes
```
grafnu: .old files necessary?

radius_server/Dockerfile.freeradius
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:18.04
```
grafnu: is there a reason not to just put etc/freeradius under radius_server/?

anurag6: No. I just thought having two etc folders might look ugly. Correcting.

topo/dva/devices_block-yellow.yaml
```diff
@@ -7,17 +7,10 @@ device_mac_placements:
...
 device_mac_behaviors:
     "9a:02:57:1e:8f:01":
         segment: INFRA
         role: "red"
     "9a:02:57:1e:8f:02":
         segment: BOS
         role: "green"
-    "9a:02:57:1e:8f:03":
-        segment: BOS
-        role: "yellow"
```
didovesei: I don't think this file needs to be updated since it is only used in test_topology.

topo/dva/behaviors_block-yellow.yaml
```diff
@@ -5,6 +5,3 @@ device_mac_behaviors:
   "9a:02:57:1e:8f:02":
     segment: BOS
     role: green
-  "9a:02:57:1e:8f:03":
-    segment: BOS
-    role: yellow
```
didovesei: I think we can modify the file name so that it makes more sense... I named it "block-yellow" because devices assigned with role "yellow" in this file will be blocked according to the ACL for that role. Since the "yellow" devices are removed, I think it is better to give the file a new name to reflect the content/test case.

topo/dva/devices_block-yellow.yaml
grafnu: I would want clarification as to why this file is changing b/c of radius testing.  My impression was that radius integration should be behavior neutral, the only difference is that the result comes from radius rather than a separate file.    If it's an adjacent cleanup of the test for some reason (e.g. it was wrong before), then please do that as a separate CL so it's crystal clear as to how all changes relate to radius integraiton tests.

topo/dva/behaviors_block-yellow.yaml
grafnu: As per other comment -- anything not required with radius integration change should be done in a separate PR.

grafnu:
> Do you think if it makes sense to compress the freeradius files into a tarball or something? I guess the files will barely change in the future but I might be wrong...

Only if there's a demonstrable reason for it.  git is good at compression, so a zip just obfuscates things. Lots of files in git isn't a problem in itself.

topo/dva/devices_block-yellow.yaml
anurag6: I shall revert this file.

topo/dva/behaviors_block-yellow.yaml
anurag6: The thing is the behaviors file is the static way of conveying behavior to faucetizer. This overwrites any result sent from the authenticator. This change would remove static behavior for faux 3 that we are testing.

etc/freeradius/certs/index.txt.attr.old
anurag6: Done.

radius_server/Dockerfile.freeradius
anurag6: Done

topo/dva/devices_block-yellow.yaml
anurag6: Done.

anurag6:
DVA test passes successfully. https://travis-ci.org/anurag6/forch/jobs/656018887?utm_medium=notification&utm_source=github_status

topo/dva/behaviors_block-yellow.yaml
grafnu: After reading both your comments more carefully I agree that the filename should be changed -- sorry, I didn't quite understand the meaning before. Also the devices_block-yellow.yaml file seems misnamed too (it actually has nothing to do with blocking yellow, that's just when it's used).

PR #48, opened 2020-02-26
Convert processes section of forch configuration to protobuf

didovesei:


proto/forch_configuration.proto
```diff
@@ -0,0 +1,25 @@
...
+ */
+syntax = "proto3";
+
+// TODO fill up sections one by one
+message ForchConfig {
+  ProcessConfig process = 1;
+}
+
+message ProcessConfig {
+  int32 scan_interval_sec = 1;
```
grafnu: we should document these since it's an "external" interface into forch, so might as well start adding the comments now.

didovesei: Added

proto/forch_configuration.proto
```diff
@@ -0,0 +1,45 @@
...
+  // scan interval in seconds
+  int32 scan_interval_sec = 1;
+
+  // indicate if Forch should check VRRP state
+  bool check_vrrp = 2;
+
+  // target processes to check
+  map<string, Process> processes = 3;
+
+  // connections to check
```
grafnu: for maps, should indicate what the key is (e.g., in this case it's "listening port")

proto/forch_configuration.proto
```diff
@@ -0,0 +1,45 @@
...
+ */
+message ProcessConfig {
+  // scan interval in seconds
+  int32 scan_interval_sec = 1;
+
+  // indicate if Forch should check VRRP state
+  bool check_vrrp = 2;
+
+  // target processes to check
+  map<string, Process> processes = 3;
```
grafnu: what's the key here?

proto/forch_configuration.proto
didovesei: Added

proto/forch_configuration.proto
didovesei: Added. It's process name

PR #52, opened 2020-02-28
Convert orchestration section in forch.yaml to use protobuf

didovesei:


forch/forchestrator.py
```diff
@@ -119,22 +119,36 @@ def initialize(self):
         self._initialized = True
 
     def _attempt_authenticator_initialise(self):
-        auth_config = self._config.get('orchestration', {}).get('auth_config')
-        if auth_config:
-            self._authenticator = Authenticator(auth_config, self.handle_auth_result)
+        orchestration_config = dict_proto(
+            self._config.get('orchestration', {}), OrchestrationConfig)
+        if not orchestration_config:
```
anurag6: NIT: We are returning for failure. It can probably be simplified.

PR #54, opened 2020-03-02
Move Forch binary entry point to forch/__main__.py

didovesei:
Mainly trying to resolve the issue where unecessary extra info is printed for "forch --version":
- Moving forch lanuch to forch/__main__.py
- Making __main__.py:main() the entry point to start forch in production environment (bin/forch is only used to start testing now)


forch/forchestrator.py
```diff
@@ -37,13 +34,17 @@
...
 
-_FORCH_CONFIG_DEFAULT = 'forch.yaml'
 _STRUCTURAL_CONFIG_DEFAULT = 'faucet.yaml'
 _BEHAVIORAL_CONFIG_DEFAULT = 'faucet.yaml'
 _SEGMENTS_VLAN_DEFAULT = 'segments-to-vlans.yaml'
 _DEFAULT_PORT = 9019
 _PROMETHEUS_HOST = '127.0.0.1'
 
+
+# TODO: 1) use protobuf after entire forch config is converted
```
grafnu: Each item should have it's own TODO (they are automatically indexed), and should be short enough to be on one line.

forch/__main__.py
```diff
@@ -0,0 +1,106 @@
...
+from forch.__version__ import __version__
+
+LOGGER = logging.getLogger('forch')
+
+_FORCH_CONFIG_DEFAULT = 'forch.yaml'
+
+
+def load_config():
+    """Load configuration from the configuration file"""
+    # TODO: 1) use protobuf after entire forch config is converted
```
grafnu: each TODO should be it's own single line

forch/__main__.py
```diff
@@ -0,0 +1,106 @@
...
+    args = parse_args(sys.argv[1:])
+
+    if args.version:
+        print(f'Forch {__version__}')
+        sys.exit()
+
+    run_forchestrator()
+
+
+if __name__ == '__main__':
```
grafnu: this seems redundant -- why define main() at all?

didovesei: bin/forch runs forch as a module so there is __main__ here.

'main()' will be run in production environment. In setup.cfg, an entry point of __main__.py:main() is added. When installing the package, a python script in bin directory will be generated, in which __main__.py:main() will be called. For the entry point settings in setup.cfg, we can only specify a function instead of a file.

forch/forchestrator.py

forch/__main__.py
didovesei: Done

didovesei:
Will merge it for next release to be in sync with changes of other components in production

PR #57, opened 2020-03-03
Built docker images

anurag6:


grafnu:
Did you push them as well? When I tried it out it's failing... We can talk through all the steps in our 1pm meeting.

PR #58, opened 2020-03-03
Add RADIUS secret handler

anurag6:


forch/authenticator.py
```diff
@@ -33,7 +33,12 @@ def __init__(self, auth_config, auth_callback=None, radius_query_object=None):
         radius_info = auth_config.radius_info
         radius_ip = radius_info.server_ip
         radius_port = radius_info.server_port
-        secret = radius_info.secret
+        secret = None
```
grafnu: better to use an if-else chain, rather than assigning multiple times. It's functionally the same, but is more "clean" in terms of understanding and refactoring. E.g.,:

if A:
  secret = f(A)
elif B:
  secrect = foo
else
  secret = None

Or consider extracting to a function, secret = self._get_radius_secret()

proto/forch_configuration.proto
```diff
@@ -94,6 +94,9 @@ message OrchestrationConfig {
 
     // Secret
     string secret = 3;
+
+    // Path to script that prints RADIUS secret
+    string radius_secret_handler = 4;
```
grafnu: helper not handler

bin/get_radius_secret
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
```
grafnu: Why do you need this file?  Can't you just do "echo SECRET"?

proto/forch_configuration.proto
grafnu: this doesn't seem to be used anywhere... should be reflected in a test

proto/forch_configuration.proto
```diff
@@ -94,6 +94,9 @@ message OrchestrationConfig {
 
     // Secret
     string secret = 3;
+
```
grafnu: plain secret should be removed

forch/authenticator.py
```diff
@@ -33,7 +33,12 @@ def __init__(self, auth_config, auth_callback=None, radius_query_object=None):
...
+        if radius_info.radius_secret_handler:
+            secret = os.popen(radius_info.radius_secret_handler).read().strip()
+        # presence of secret field overrides any secret retreived from handler
+        if radius_info.secret:
```
grafnu: this should be removed -- there never should be an explicit secret in the config file -- no need to specify two ways of doing the same thing.

bin/get_radius_secret
anurag6: as in the helper file itself?

grafnu:
Yes

anurag6:
I built the docker images and uploaded them to docker hub. This should be the final draft, I have addressed the previous comments. PTAL @grafnu

forch/authenticator.py
```diff
@@ -148,7 +151,7 @@ def parse_args(raw_args):
                         help='RADIUS server ip')
     parser.add_argument('-p', '--server-port', type=int, default=1812,
                         help='Server port that freeradius is listening on')
-    parser.add_argument('-r', '--radius-secret', type=str, default='SECRET',
+    parser.add_argument('-r', '--radius-secret', type=str, default='echo SECRET',
```
grafnu: somethings wrong, "echo SECRET" is not the same as the radius_secret itself. I think in this case (command line arg), it should be the secret, and the default value would be SECRET (without echo). No need to have the helper for a command line argument.

anurag6: But the command line arguments essentially builds the config data structure. This was why I had left the 'secret' field in place previously. Authenticator is fed this information through the config file. We either bring back the secret field or change the name of the command line argument,

grafnu:
k -- needs to be consistent, but don't bring back the secret!

forch/authenticator.py
anurag6: Updated. PTAL.

anurag6:
> nb: this could have been f'echo {ARGS.radius_secret}' as per the previous discussion. Change or ignore, your call!

Makes sense.

PR #61, opened 2020-03-10
Add scripts that set up Faucet and Forch on debain machine

didovesei:


bin/setup_controller
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash -e
+
+setup_base=
+faucet_port=6653
+prom_port=9302
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        -s|--setup_base)
+            skip_docker_build=y
```
grafnu: this doesn't seem right... setup_base is always empty, and this flag then doesn't build docker?

topo/controller/faucet_structural.yaml
```diff
@@ -0,0 +1,45 @@
+dps:
+  sw1:
+    dp_id: 0x12345
+    interfaces:
+      1:
```
grafnu: These interface assignemnts don't make any sense to me, but that's no a big deal since it's easy to tweak later.  I don't understand how this is mapping out to what it would look like on a physical switch (if that was somewhat the intent).

bin/setup_controller
didovesei: Done

topo/controller/faucet_structural.yaml
didovesei: Removed interface 7. Tested on switch and didn't see problems.

didovesei:
> Also would be good to get a simple integration test in somewhere for this... since otherwise it'll be very fragile to any changes. Just enough to call the setup script and do a ping between two devices.

Added

topo/solo/faucet_structural.yaml
```diff
@@ -0,0 +1,38 @@
...
+        tagged_vlans: [100,200,300,400]
+      4:
+        description: "IoT-2"
+        native_vlan: 100
+        max_hosts: 1
+      5:
+        description: "egress-3"
+        tagged_vlans: [100,200,300,400]
+      6:
+        description: "host"
```
grafnu: Nit: Why is this "host" while the others are "IoT-2"?

PR #63, opened 2020-03-16
Make faucetizer output behavior configurable

didovesei:


forch/forchestrator.py
```diff
@@ -96,6 +99,9 @@ def initialize(self):
         self._process_static_device_behavior()
         if self._faucetizer:
             faucetizer.write_behavioral_config(self._faucetizer, self._behavioral_config_file)
+            faucet_config_handler = FaucetConfigFileHandler(
+                self._faucetizer, self._structural_config_file, self._behavioral_config_file)
+            self._file_observer.schedule(faucet_config_handler, os.getenv('FAUCET_CONFIG_DIR'))
```
grafnu: I would push the Observer down into file_handlers -- there's no good reason to expose that part of the abstraction here.

forch/file_handlers.py
```diff
@@ -0,0 +1,25 @@
...
+        self._behavior_config_file = behavior_config_file
+
+    def on_modified(self, event):
+        """on file modified"""
+        super(FaucetConfigFileHandler, self).on_modified(event)
+
+        if event.is_directory or self._structural_config_file != event.src_path:
+            return
+
+        forch.faucetizer.update_structural_config(self._faucetizer, self._structural_config_file)
```
grafnu: this is duplicated code/logic with the current periodic mechanism. I think it's good to keep the old functionality for now (remove in another pass if necessary), just in case this new stuff doesn't work in production, but this logic (what happens when something changes) should be captured in one place somewhere.

forch/file_handlers.py
```diff
@@ -0,0 +1,25 @@
+"""File changes watcher"""
```
grafnu: It's not clear what abstraction you're going for here... Is this file meant to be specific to Faucet config files (has faucet-specific file stuff in it), or generic for watching changes in any file?

forch/forchestrator.py
```diff
@@ -9,6 +9,7 @@
 
 from google.protobuf.message import Message
 from forch.proto import faucet_event_pb2 as FaucetEvent
+from watchdog.observers import Observer
```
grafnu: Does this debian package already exist in rapture?  We should check that asap b/c if not, then I'm not sure we want to be pulling in that dependency (it's a lot of work).

forch/forchestrator.py
didovesei: Created a FileWatcher class in file_watchers.py

forch/file_handlers.py
didovesei: Ok, now it is still periodically writing out behavioral config, while updating structural config is triggered by file change.

forch/file_handlers.py
didovesei: It is specific to faucet config files but we may also put other file handlers in this file.

forch/forchestrator.py
didovesei: Yes, there is python3-watchdog in rapture.

forch/file_handlers.py
grafnu: I don't understand what your point is. It's not a question of when the thing is triggered, it's about what it does when it's triggered...?

forch/file_handlers.py
grafnu: I think you should commit either way -- straddling the fence like this just leads to spagetti since you don't have clean abstraction either way. If a more generic something is needed later, it's better to do the refactor as needed. So -- should either be a generic thing while the faucet-specific stuff is in forchestrator.py, or a faucet-specific thing...

didovesei:
Made faucetizer output behavior configurable such that if faucetize_interval is unset or set to 0, faucetizer immediately writes behavioral config after an inode (device placement and behavior, structural config) change; otherwise it periodically writes out behavioral config.

forch/faucetizer.py
```diff
@@ -18,14 +18,32 @@
...
         self._structural_faucet_config = structural_faucet_config
         self._behavioral_faucet_config = None
-        self._lock = threading.Lock()
-
+        self._config = orch_config
+        self._behavioral_config_file = behavioral_config_file
+        self._lock = threading.RLock()
+
+    # pylint: disable=no-self-argument, protected-access, no-method-argument
+    def _write_behavioral_config():
```
grafnu: this is not a good use of annotations -- it's taking something that can/should be very simple and adding an unnecessary layer of abstraction on top of it -- makes it very hard to read/understand the code. This really should just be a regular function that's called as needed.

didovesei: Ok, removed annotations and created _write_behavioral_config to wrap the oeprations.

proto/forch_configuration.proto
```diff
@@ -64,6 +64,9 @@ message OrchestrationConfig {
   string segments_vlans_file = 5;
 
   // interval of periodic faucetizing task in seconds
+  // NB: not setting the interval or explicitly setting it to 0
```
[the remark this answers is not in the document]
didovesei: Rephrased..

forch/faucetizer.py
```diff
@@ -60,7 +45,9 @@ def process_device_placement(self, eth_src, placement, static=False):
                 if removed:
                     LOGGER.info('Removed %s device: %s', device_type, eth_src)
 
-    @_write_behavioral_config()
+            if not self._config.faucetize_interval_sec:
```
grafnu: if condition should be moved into a function rather than replicating 3x, and then rename function something like "_flush_behavorial_config()" -- since it's not strictly always writing it.  If you need both behaviors, then you could either 1) create another intermediate function, or 2) add an optional force=True argument for the cases when you always want to write it.

forch/faucetizer.py
```diff
@@ -105,7 +103,9 @@ def _faucetize(self):
 
         self._behavioral_faucet_config = behavioral_faucet_config
 
-    def _write_behavioral_config(self):
+    def _flush_behavioral_config(self, force=False):
+        if not force and self._config.faucetize_interval_sec:
+            return
```
grafnu: this method still overlaps with write_behavioral_config, so should be consolidated. Also write_behavioral_config should be class method, not a static one.  The objective here is to consolidate as much common stuff in one place as possible -- so the yaml.dump method should only in one place, and then have other functions that provide other functionality on top of that. In the long run, this minimizes mistakes since there's only one place something happens, so you don't need to chase down and/or edit multiple instances of something.

didovesei: OK, moved both write_behavioral_config and update_structural_config into Faucetizer as class methods.

forch/faucetizer.py
```diff
@@ -147,16 +150,7 @@ def load_faucet_config(file):
...
-        structural_config = yaml.safe_load(structural_config_file)
-        faucetizer.process_faucet_config(structural_config)
-
-
-def write_behavioral_config(faucetizer: Faucetizer, file):
-    """Get behavioral config from faucetizer and write to file"""
-    behavioral_config = faucetizer.get_behavioral_faucet_config()
-    with open(file, 'w') as behavioral_config_file:
-        yaml.dump(behavioral_config, behavioral_config_file)
+    pass
```
grafnu: why not just delete the entire method?

didovesei: Forgot to commit.. Deleted.

PR #64, opened 2020-03-16
Forch varz interface with basic varz

anurag6:


grafnu:
Looks good overall -- should also get a minimal unit test in here, just enough to validate that it's working.

debian/control
```diff
@@ -26,6 +26,7 @@ Architecture: all
 Depends: python3-faucet (= FAUCET_VERSION),
          python3-netaddr,
          python3-networkx,
+         python3-prometheus-client (>= 0.7.1), python3-prometheus-client (<< 0.7.2),
```
grafnu: I think you can just do = for exact match

forch/forch_metrics.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def stop(self):
+        """Kill varz server"""
+        self._http_server.stop_server()
+
+    def update_var(self, var, value):
+        """Update given varz with new value"""
+        varz = self._metrics.get(var)
+        if not varz:
+            LOGGER.error('Error updating to varz %s since it is not known.', var)
```
grafnu: seems like this should be a hard error (throw exception)

forch/forch_metrics.py
```diff
@@ -0,0 +1,63 @@
...
+        if isinstance(varz, Info):
+            varz.info(value)
+        else:
+            LOGGER.error('Error updating to varz %s since it\'s type %s is not known.',
+                         var, type(varz))
+            return
+
+    def _add_var(self, var, var_help, metric_type):
+        """Add varz to be tracked"""
+        self._metrics[var] = metric_type(var, var_help, registry=self._reg)
```
grafnu: again hard error

forch/forchestrator.py
```diff
@@ -99,6 +100,8 @@ def initialize(self):
 
         self._validate_config_files()
 
+        self._forch_metrics = ForchMetrics(8302, self._config.http)
```
grafnu: this is not right -- there should only be one config block for this, and the port should be part of it.  I think we might need to do an audit of the config setup -- The current http config makes no sense for this (it's specifying a file root).  If you want to simplify and not make things worse, find a way to do this without passing self._config.http in.

anurag6: Makes sense. Let me look into it.

.travis.yml
```diff
@@ -9,13 +9,11 @@ addons:
   apt:
     update: true
```
anurag6: Will remove once travis tests go through

forch/forch_metrics.py
anurag6: Done

forch/forch_metrics.py
anurag6: Done.

forch/forchestrator.py
anurag6: Changed now.

debian/control
anurag6: Done

anurag6:
@grafnu  PTAL.

debian/control
```diff
@@ -26,6 +26,7 @@ Architecture: all
 Depends: python3-faucet (= FAUCET_VERSION),
          python3-netaddr,
          python3-networkx,
+         python3-prometheus-client (= 0.7.1), python3-prometheus-client (<< 0.7.2),
```
grafnu: why << 0.7.2?

forch/forchestrator.py
```diff
@@ -99,6 +100,8 @@ def initialize(self):
 
         self._validate_config_files()
 
+        self._forch_metrics = ForchMetrics(8302)
```
grafnu: port should be configurable

debian/control
anurag6: Changed

forch/forchestrator.py
anurag6: Done.

debian/control
```diff
@@ -26,7 +26,7 @@ Architecture: all
 Depends: python3-faucet (= FAUCET_VERSION),
          python3-netaddr,
          python3-networkx,
-         python3-prometheus-client (= 0.7.1), python3-prometheus-client (<< 0.7.2),
+         python3-prometheus-client (= 0.7.1),,
```
grafnu: extra comma

forch/http_server.py
```diff
@@ -78,7 +78,10 @@ def __init__(self, port, config=None):
         self._config = config
         self._paths = {}
         self._server = None
-        self._root_path = config.http_root if config else 'public'
+        if config and config.http_root:
```
grafnu: self._root_path = config.http_root if config and config.http_root else 'public'

proto/forch_configuration.proto
```diff
@@ -150,3 +153,11 @@ message EventClientConfig {
   // port debounce time in seconds
   int32 port_debounce_sec = 1;
 }
+
+/*
+ * Varz interface section in Forch configuration
+ */
+message VarzInterfaceConfig {
+  // port debounce time in seconds
```
grafnu: ?

forch/forch_metrics.py
```diff
@@ -8,14 +8,16 @@
...
 
 class ForchMetrics():
     """Class that implements the module that exposes varz for metrics"""
     _reg = REGISTRY
 
-    def __init__(self, local_port):
-        self._local_port = local_port
+    def __init__(self, varz_config):
+        self._local_port = varz_config.varz_port or DEFAULT_VARZ_PORT
+        LOGGER.info('Anurag forch_metrics port is %s', self._local_port)
```
grafnu: reminder to remove Anurag (I'd still leave the logging statement)

anurag6: Yeah my bad.

debian/control
anurag6: Done

forch/http_server.py
anurag6: Done.

proto/forch_configuration.proto
anurag6: Done.

forch/forch_metrics.py
anurag6: Done.

PR #67, opened 2020-03-24
Remove auth.yaml dependency

anurag6:


anurag6:
Minor change to get rid of auth.yaml dependency since it doesnt do anything and only hinders authenticator's use in standalone mode.

PR #72, opened 2020-04-01
Check both radius conditions in tests

grafnu:
I didn't want to block your PR, but this is more the refactoring I had in mind. The main reason is that it's good to check both cases (received and timeout) both times, just to make sure the case where "both" happens doesn't pass.  The differentiation of the results should primarily be in the expected output, not the mechanism itself -- this way, both calls to "check varz" are the same, it's just the output that's different/matched.


anurag6:
Makes sense. But I was planning to use the same method to test for other varz. Well probably I can write another method to sumply print the value of the varz. Yup. LGTM.

PR #73, opened 2020-04-03
Automatically assign cookies to ACL rules

didovesei:


forch/config_file_watcher.py
```diff
@@ -0,0 +1,98 @@
...
+
+from watchdog.observers import Observer
+from watchdog.events import FileSystemEventHandler
+
+LOGGER = logging.getLogger('watcher')
+
+
+class ConfigFileWatcher:
+    """Watch file changes"""
+    def __init__(self, structural_config_file, faucetizer):
```
grafnu, line 16: this is suffering from dependency inversion -- this class shouldn't know anything about "faucetizer" or faucetizer specific logic (the _reschedule_acl_file_handlers function). Rather, it should accept lambda functions that perform the callback, and then the faucetizer specific logic should be handled outside of this class.

forch/config_file_watcher.py
```diff
@@ -0,0 +1,98 @@
...
+        self._acl_watches = []
+        with open(self._structural_config_file) as file:
+            structural_config = yaml.safe_load(file)
+            for acl_config_file in structural_config.get('include', []):
+                acl_config_handler = AclConfigFileHandler(self._faucetizer, acl_config_file)
+                watch = self._observer.schedule(acl_config_handler, self._path)
+                self._acl_watches.append(watch)
+
+
+class ConfigFileHandler(FileSystemEventHandler):
```
grafnu, line 48: this function looks fine since there's no faucetizer-specific logic

forch/config_file_watcher.py
```diff
@@ -0,0 +1,98 @@
...
+
+            self._last_hash = hash
+
+        self._on_modified()
+
+    def _on_modified(self):
+        pass
+
+
+class FaucetConfigFileHandler(ConfigFileHandler):
```
grafnu, line 77: this should be inverted so they're not here, but rather closer to where they're "used" -- e.g. in the faucetizer module since they are faucetizer specific.

forch/config_file_watcher.py
```diff
@@ -0,0 +1,98 @@
...
+    """Handles Faucet config file changes"""
+    def __init__(self, faucetizer, structural_config_file, structural_config_handler):
+        super(FaucetConfigFileHandler, self).__init__(structural_config_file)
+        self._faucetizer = faucetizer
+        self._structural_config_handler = structural_config_handler
+
+    def _on_modified(self):
```
grafnu, line 84: I think you can likely make this cleaner using a lambda function for "on_modified" rather than having an sntire class. The init method of both of these classes don't do anything interesting, and they only have logic in one function -- so much cleaner to use a lambda function for that one method rather than class inheritence.

forch/faucetizer.py
```diff
@@ -26,10 +28,12 @@ def __init__(self, orch_config, structural_config_file, segments_to_vlans,
         self._segments_to_vlans = segments_to_vlans
         self._structural_faucet_config = None
         self._behavioral_faucet_config = None
+        self._acl_configs = {}
+        self._next_cookie = 1
```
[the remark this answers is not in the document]
didovesei: initialized to None

forch/faucetizer.py
```diff
@@ -119,6 +149,14 @@ def flush_behavioral_config(self, force=False):
         with open(self._behavioral_config_file, 'w') as file:
             yaml.dump(self._behavioral_faucet_config, file)
 
+    def flush_acl_config(self, file_name, acls_config, force=False):
+        """Generate and write acl configs to files"""
+        if not force and self._config.faucetize_interval_sec:
+            return
+        with open(file_name, 'w') as acl_file:
+            yaml.dump(acls_config, acl_file)
+
```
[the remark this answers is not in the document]
didovesei: Done

forch/config_file_watcher.py
didovesei: Moved _reschedule_acl_file_handlers to forchestrator; passing faucetizer logic as callback

forch/config_file_watcher.py
didovesei: Removed. Logic is passed as callback now.

forch/config_file_watcher.py
didovesei: Removed. Passing faucetizer's method as callback.

forch/config_file_watcher.py
```diff
@@ -13,16 +12,14 @@
 
 class ConfigFileWatcher:
     """Watch file changes"""
-    def __init__(self, structural_config_file, faucetizer):
+    def __init__(self, structural_config_file, structural_config_modified_callback):
```
grafnu, line 15: there is no need to restrict this to "structural" here -- you're presuming the use of a (generic) class. You can remove "structural" in here (all places) and it's still accurate and more general-purpose. Realistically, nothing in here actually has anything to do with a config file either, correct?  It's really just a "FileChangeWatcher", so you might consider making it more generic than it is.

forch/config_file_watcher.py
```diff
@@ -32,67 +29,42 @@ def stop(self):
         """Stop watcher"""
         self._observer.stop()
 
-    def _reschedule_acl_file_handlers(self):
+    def schedule_acl_file_handler(self, acl_file_name, on_modified_callback):
```
grafnu, line 32: this function is unnecessarily tied to an "acl" -- what does it actually have to do with the operation of the function? it's really just "add_file_watch_handler" or something like that, correct?

forch/faucetizer.py
```diff
@@ -21,20 +21,20 @@
...
         self._segments_to_vlans = segments_to_vlans
         self._structural_faucet_config = None
         self._behavioral_faucet_config = None
-        self._acl_configs = {}
-        self._next_cookie = 1
+        self._next_cookie = None
         self._config = orch_config
         self._structural_config_file = structural_config_file
         self._behavioral_config_file = behavioral_config_file
+        self._reschedule_acl_file_handlers = reschedule_acl_file_handlers
```
grafnu, line 36: the use of the word "reschedule" is very strange here.  What are you trying to say?  Make sure you use a descriptive word of what it is.  I think part of the problem is you're perpetuating the use of the term "schedule" from the observer class... which makes no sense (there is no schedule involved).  So, I wouldn't perpetuate their bad terminology outside of the file watcher class.

topo/dva/faucet_block-yellow.yaml
```diff
@@ -692,8 +692,8 @@ dps:
       max_per_interval: 5
       send_interval: 5
 include:
-- uniform-acls.yaml
-- role-based-acls.yaml
+- uniform-acls_augmented.yaml
```
grafnu, line 695: what is this file, and what is this change for?  I can't tell just from this context and it's not obvious from the code how this file is used. I just want to make sure the abstraction isn't bleeding out inappropriatley.

didovesei: So Faucetizer reads ACLs from included ACL files (uniform-acls.yaml and role-based-acls.yaml), adds cookies to the ACLs and then writes the ACLs to other files (uniform-acls_augmented.yaml and role-based-acls_augmented.yaml), so that it does not overwrite the existing .yaml files, similar to faucet_structural.yaml and faucet.yaml.

forch/config_file_watcher.py
didovesei: So I think it has to distinguish between the main faucet_structural.yaml and included files. There may be a case where included files are changed in faucet_structural.yaml. In this the handlers for the old included files should be unscheduled, but the handler for faucet_structural.yaml the handler should remain unchanged.

forch/config_file_watcher.py
didovesei: So the reason is the same with the comment above. What I wanted to do is to only store the scheduled watches for the ACL files but not faucet_structural.yaml, since the include section may change in faucet_structural.yaml.

forch/faucetizer.py
didovesei: This function is supposed to unregister existing handlers for all ACL files and register new handlers for new ACL files. Maybe refresh_acl_file_handlers?

topo/dva/faucet_block-yellow.yaml
grafnu: What is the purpose of the "faucet_block-yellow.yaml" file, how is that used?  Why is this change reflected in source-control... seems like the mechanism you describe should be internal only.

forch/faucetizer.py
grafnu: reregister?

forch/config_file_watcher.py
grafnu: what does that have to do with the logic of a config file handler?  The root problem is you're conflating the concept of a "file change watcher" and "structural vs. behavior config files"

forch/config_file_watcher.py
grafnu: why can't this just be about adding a generic file watch? Why does it have to be acl files?

topo/dva/faucet_block-yellow.yaml
grafnu: why is this change in source control, though? The augmentation you describe should be internal and not exposed. Why is this change in faucet_block-yellow.yaml?  When is this file (faucet_block-yellow.yaml) used?

forch/faucetizer.py
grafnu: reregister_ or refresh_ would be fine

topo/dva/faucet_block-yellow.yaml
didovesei: That file is used for test_topology. Basically the testing script will compare generated file with that file and tests if they are identical.

forch/config_file_watcher.py
didovesei: OK will make it a generic file watcher and move other logic to forchestrator

grafnu:
Got it, thanks. I wanted to make sure it was for a unit test (which it is),
and not a system integration test.

forch/faucetizer.py
didovesei: Using _reregister_acl_file_handlers now.

forch/config_file_watcher.py
didovesei: Making it a generic FileChangeWatcher now. 

Also restructured the watcher and the handler class because it seems that the low level Inotify library on Linux system only supports monitoring a directory and not an individual file.

forch/faucetizer.py
```diff
@@ -69,13 +74,50 @@ def process_device_behavior(self, eth_src, behavior, static=False):
...
+
+            if not self._config.faucetize_interval_sec and self._reregister_acl_file_handlers:
+                self._reregister_acl_file_handlers(self._watched_acl_files, new_watched_acl_files)
+            self._watched_acl_files = new_watched_acl_files
+
             self.flush_behavioral_config()
 
+    def _process_acl_config(self, file_path, acls_config):
+        new_acls_config = copy.copy(acls_config)
+        if not self._next_cookie:
```
grafnu, line 102: why is this necessary?  Under what condition is _next_cookie not defined?

didovesei: With current code it is always set to 1 before being used... Just wanted to be able to notice corner cases during run time. Removing this check..

grafnu:
Sure -- but if you want to notice corner cases it's better to fail hard
with an exception. Otherwise, it can be really hard to trace down problems
(ends up being a needle in a haystack once it's out there in the wild!)

didovesei:
> Sure -- but if you want to notice corner cases it's better to fail hard with an exception. Otherwise, it can be really hard to trace down problems (ends up being a needle in a haystack once it's out there in the wild!)

Right, raising an exception now. I think the PR is ready for a final review... PTAL.

PR #74, opened 2020-04-07
Disable tests. Give prometheus-client version a range

anurag6:


bin/test_basics
```diff
@@ -4,7 +4,7 @@ source bin/stack_functions
 
 preamble_setup
 
-bin/package_forch testing
+#bin/package_forch testing
```
grafnu, line 7: bug reference

debian/control
```diff
@@ -26,7 +26,7 @@ Architecture: all
 Depends: python3-faucet (= FAUCET_VERSION),
          python3-netaddr,
          python3-networkx,
-         python3-prometheus-client (= 0.7.1),
+         python3-prometheus-client (>= 0.7.1), python3-prometheus-client (<< 0.7.2),
```
grafnu, line 29: What does this do?  How is this different than fixing the exact version?  What does << mean... is that functionally <=?  What are the versions that this is trying to match?  Not really a problem with the PR, just trying to understand the debian semantics.

anurag6: So rapture has versions like 0.7.1-1 and 0.7.1-1.1. = 0.7.1 would fail. << is functionally same as less than is what I believe. Basically exclude 0.7.2

bin/test_basics
anurag6: Addressed.

grafnu: you should probably put the right bug number in there.

anurag6: Changed.

PR #79, opened 2020-04-15
Set egress state to unkown if no LAG info is received

didovesei:
Set egress state to unkown in NOAH API output if no LAG info is received

### DIFF
--- a/public/protos.html
+++ b/public/protos.html
@@ -4373,56 +4373,62 @@ or to host_path API to list path </p></td>
           <tbody>
             
               <tr>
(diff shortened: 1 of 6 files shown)


public/protos.html
```diff
@@ -4373,56 +4373,62 @@ <h3 id="State.State">State.State</h3>
           <tbody>
             
               <tr>
-                <td>broken</td>
+                <td>unkown</td>
```
pbatta, line 4383: s/unkown/unknown ?

PR #82, opened 2020-04-20
WIP: Change topo/bond/faucet.yaml to replace interface_ranges and raise a hard exception when port is not defined in config.

anurag6:


forch/faucet_state_collector.py
```diff
@@ -607,8 +607,7 @@ def _fill_port_behavior(self, switch_name, port_id, port_map, metrics=None):
 
         port_config = dp_config.ports.get(port_id)
         if not port_config:
-            LOGGER.warning('Port not defined in dps config: %s, %s', switch_name, port_id)
-            return
+            raise Exception('Port not defined in dps config: %s, %s', switch_name, port_id)
```
grafnu, line 610: would be good to (also) have an explicit check for "interface_ranges not allowed" -- right now the error message doesn't indicate the actual problem.

anurag6: Addressed. PATL

PR #84, opened 2020-04-23
Fix egress state traceback error

grafnu:


forch/faucet_state_collector.py
```diff
@@ -433,7 +433,8 @@ def _fill_egress_state(self, target_obj):
                 target_obj[EGRESS_STATE] = State.unknown
                 target_obj[EGRESS_DETAIL] = 'No LAG information received'
             else:
-                target_obj[EGRESS_STATE] = State.State.Name(egress_obj.get(EGRESS_STATE))
+                egress_state = egress_obj.get(EGRESS_STATE, State.unknown)
```
didovesei, line 436: This seems to be covered by the same check in line 432?

I mean, in line 432, if `EGRESS_STATE` is not in `egress_obj`, then it sets `target_obj[EGRESS_STATE]` to `State.unkown`. Here, it does similiar thing, which sets `egress_state` to `State.unkown`.

Also what's the difference between `State.unkown` and `State.State.Name(State.unkown)`? Can we just set `target_obj[EGRESS_STATE]` to `State.unkown`?

grafnu: You are right about the "this is already taken care of"... I'll have to have a think about that. It might be that this bug was already fixed.  I'm not sure if it's still happening or not.

State.unknown is a numerical constant, 0 -- while State.State.Name(State.unknown) is the string 'unknown' -- so it just depends on what's wanted in this case. At the very least they should be consistent.

PR #85, opened 2020-04-23
Proxy server to serve up varz on various ports

anurag6:


proto/forch_configuration.proto
```diff
@@ -165,6 +165,15 @@ message EventClientConfig {
  * Varz interface section in Forch configuration
  */
 message VarzInterfaceConfig {
-  // port number to curl for varz list
-  int32 varz_port = 1;
+  // port number to curl for forch varz list
+  int32 forch_varz_port = 1;
```
grafnu, line 169: this should be a generic map configuraiton, not specific to forch/faucet/gauge.  E.e., the config should have something like:

proxy_requests:
  proxy_port: 8032
  targets:
    faucet:
      port: 9032
    forch:
      port: 9036

forch/forch_metrics.py
```diff
@@ -85,6 +107,31 @@ def get_metrics(self, path, params):
...
         return f"Error creating varz interface: {str(error)}"
+
+    def _get_url(self, server, port):
+        return 'http://' + str(server) + ':' + str(port)
+
+    def _register_metric_page(self, path, server, port):
+        self._metric_pages[path] = self._get_url(server, port)
+
+    def _register_metric_pages(self):
+        self._register_metric_page('faucet', LOCALHOST,
```
grafnu, line 129: this should all be dynamicallky pulled from config, not hard-coded

proto/forch_configuration.proto
```diff
@@ -165,6 +165,15 @@ message EventClientConfig {
...
+
+  // port number to curl for faucet varz list
+  int32 faucet_varz_port = 2;
+
+  // port number to curl for gauge varz list
+  int32 gauge_varz_port = 3;
+
+  //port number for varz proxy
+  int32 proxy_varz_port = 4;
```
grafnu, line 178: nothing about this proxy is specific to varz, so it shouldn't be encoded in the names

forch/forch_metrics.py
```diff
@@ -32,6 +41,19 @@ def start(self):
         finally:
             threading.Thread(target=self._http_server.start_server(), daemon=True).start()
 
+        self._register_metric_pages()
+
+        self._proxy_server = forch.http_server.HttpServer(self._proxy_port)
+        try:
+            for page in self._metric_pages:
+                self._proxy_server.map_request(
+                    page, lambda path, params: self._get_path_metrics(path.split('/')[1]))
```
grafnu, line 50: what about the rest of the path?  Shouldn't this just be everything except the first bit?

anurag6: I didnt get it. Rest of the path?

grafnu: You take the input path, split it by /, and then only take [1] -- what about [2].. [n]?

forch/forch_metrics.py
```diff
@@ -85,6 +107,31 @@ def get_metrics(self, path, params):
         """Return metric list in printable form"""
         return generate_latest(self._reg).decode('utf-8')
 
+    def _get_path_metrics(self, path):
```
grafnu, line 110: also, just realized you're perpetuating the wrong abstraction here.... this is just a web proxy, it has nothing to do with metrics... so likely should be pulled out into its own separate class.

forch/forch_metrics.py
anurag6: Addressed.

proto/forch_configuration.proto
anurag6: Addressed.

proto/forch_configuration.proto
anurag6: Addressed.

forch/forch_metrics.py
anurag6: Addressed.

forch/forch_metrics.py
anurag6: Makes sense. Created new file.

anurag6:
Made changes. PTAL @grafnu

bin/check_style
```diff
@@ -12,7 +12,7 @@ if [ -z "$VIRTUAL_ENV" -a -d venv ]; then
     source venv/bin/activate
 fi
 
-PY_LIB=$VIRTUAL_ENV/lib/python3.7/site-packages/
+PY_LIB=$VIRTUAL_ENV/lib/python3.6/site-packages/
```
grafnu, line 15: we've standardized on 3.7

forch/forch_proxy.py
```diff
@@ -0,0 +1,71 @@
...
+    """Class that implements the module that creates a proxy server"""
+
+    def __init__(self, proxy_config):
+        self._proxy_config = proxy_config
+        self._proxy_port = self._proxy_config.proxy_port or DEFAULT_PROXY_PORT
+        self._pages = {}
+        self._proxy_server = None
+
+    def start(self):
+        """Start serving varz"""
```
grafnu, line 25: this class should just be a generic web proxy, so there should be no mention of varz in this file...

forch/forchestrator.py
```diff
@@ -128,6 +130,8 @@ def initialize(self):
         self._process_static_device_behavior()
         if self._faucetizer:
             self._faucetizer.flush_behavioral_config(force=True)
+        self._proxy = ForchProxy(self._config.proxy_server)
```
grafnu, line 133: ForchProxy should only be instantiated/started if specifically enabled. So, you could either trigger it off of the port being defied and do away with DEFAULT_PROXY_PORT as a default, or trigger it off of the proxy config block being defined.  Either way, an empty config file should not instantiate the proxy at all.

proto/forch_configuration.proto
```diff
@@ -165,6 +168,28 @@ message EventClientConfig {
  * Varz interface section in Forch configuration
  */
 message VarzInterfaceConfig {
-  // port number to curl for varz list
+  // port number for forch varz
   int32 varz_port = 1;
 }
+
+/*
+ * Target to scrape metrics from
```
grafnu, line 176: it's a target for a port proxy, has nothing to do with metrics -- names/comments here and below should reflect that.

proto/forch_configuration.proto
```diff
@@ -165,6 +168,28 @@ message EventClientConfig {
...
+ */
+message ProxyServerConfig {
+  // port number for varz proxy
+  int32 proxy_port = 1;
+
+  // list of targets to scrape metrics from
+  repeated MetricTarget targets = 2;
```
grafnu, line 183: this should be a map of names that point at targets. It's nonsensical to have two names that are the same, and maps are strictly better than lists for sorting information unless you need to preserve ordering or allow duplicated.  See ProcessConfig for an example.  (Note that the value of the map should still be an object, even if it only has one item in it... as future-proofing)

topo/bond/forch.yaml
```diff
@@ -9,6 +9,14 @@ site:
       port: 9020
 orchestration:
   static_device_placement: placement.yaml
+proxy_server:
```
grafnu, line 12: also wouuld expect an update/change to the integration testing. It should be nearly transparent, just need to update the URL somewhere for what it's pointing at to test a raw varz access (which I believe is there somewhere).  Should be easy to check both proxied and raw access.

bin/check_style
anurag6: My mistake. Addressed.

forch/forch_proxy.py
anurag6: Addressed.

proto/forch_configuration.proto
anurag6: Addressed.

proto/forch_configuration.proto
anurag6: Addressed.

topo/bond/forch.yaml
anurag6: Yup. Currently added to test_sol. Will replace other tests.

forch/forchestrator.py
anurag6: Addressed

anurag6:
@grfnu PTAL

bin/stack_functions
```diff
@@ -400,6 +400,11 @@ function validate_radius_varz {
 
 function test_varz_value {
     varz=$1
-    value=$( printf "%.0f" $(wget 0.0.0.0:8302 -O- | grep $varz | grep -v 'HELP\|TYPE' | awk '{print $2'}) )
+    if [ -z "$2" ]; then
+        path=8080/$2
```
grafnu, line 404: not a path -- it's port + path ?

How does this possibly work?  -z tests for the value being zero?  I think you need a negative test somewhere.

bin/test_stack
```diff
@@ -79,10 +79,10 @@ sudo ip link set t1sw2-eth28 up
...
 sleep 32
 test_forch -healthy
-test_varz_value process_state{process=\"faucet\"}
-test_varz_value process_state{process=\"gauge\"}
-test_varz_value process_state{process=\"forch\"}
-test_varz_value process_state{process=\"sleep\"}
+test_varz_value process_state{process=\"faucet\"} forch
+test_varz_value process_state{process=\"gauge\"} forch
+test_varz_value process_state{process=\"forch\"} forch
+test_varz_value process_state{process=\"sleep\"} forch
```
grafnu, line 85: make one of this native (empty), and then also add in a negative failing case (e.g. xyzzy)

proto/forch_configuration.proto
```diff
@@ -173,23 +173,20 @@ message VarzInterfaceConfig {
...
+ * Proxy server config
  */
 message ProxyServerConfig {
-  // port number for varz proxy
+  // port number for proxy server
   int32 proxy_port = 1;
 
-  // list of targets to scrape metrics from
-  repeated MetricTarget targets = 2;
+  // map of targets
```
grafnu, line 182: indicate what they map key is

proto/forch_configuration.proto
```diff
@@ -179,7 +179,7 @@ message ProxyServerConfig {
   // port number for proxy server
   int32 proxy_port = 1;
 
-  // map of targets
+  // map of targets with target name as key
```
grafnu, line 182: "target with target name" is a bit of a circular definition. It's a path prefix as key, right?

bin/stack_functions
anurag6: Addressed.

bin/test_stack
anurag6: Addressed.

proto/forch_configuration.proto
anurag6: Addressed.

proto/forch_configuration.proto
anurag6: Addressed.

anurag6:
Addressed concerns. @grafnu  PTAL

.travis.yml
```diff
@@ -9,13 +9,11 @@ addons:
   apt:
     update: true
 install:
-  - set -e
```
grafnu, line 12: restore

bin/stack_functions
```diff
@@ -400,11 +400,13 @@ function validate_radius_varz {
 
 function test_varz_value {
     varz=$1
-    if [ -z "$2" ]; then
-        port_path=8080/$2
+    port=$2
+    path=$3
+    if [ -z "$3" ]; then
```
grafnu, line 405: use $path here

bin/stack_functions
```diff
@@ -400,11 +400,13 @@ function validate_radius_varz {
...
+        query_path=0.0.0.0:$port
     else
-        port_path=8302
+        query_path=0.0.0.0:$port/$path
```
grafnu, line 408: If statement is likely unnecessary, since 0.0.0.0:8080 is equivalent to 0.0.0.0:8080/

bin/test_stack
```diff
@@ -79,10 +79,11 @@ sudo ip link set t1sw2-eth28 up
...
 test_forch -healthy
-test_varz_value process_state{process=\"faucet\"} forch
-test_varz_value process_state{process=\"gauge\"} forch
-test_varz_value process_state{process=\"forch\"} forch
-test_varz_value process_state{process=\"sleep\"} forch
+test_varz_value process_state{process=\"faucet\"} 8302
+test_varz_value process_state{process=\"gauge\"} 8080 forch
+test_varz_value process_state{process=\"forch\"} 8080 forch
+test_varz_value process_state{process=\"sleep\"} 8080 forch
+test_varz_value process_state{process=\"sleep\"} 8080 random_string
```
grafnu, line 86: https://xkcd.com/221/ -- not really random :-) -- better use something like unknown or error -- the intention behind it

.travis.yml
anurag6: Addressed.

bin/stack_functions
anurag6: Obsolete.

bin/stack_functions
anurag6: Addressed.

bin/test_stack
anurag6: Addressed

bin/test_stack
```diff
@@ -83,7 +83,7 @@ test_varz_value process_state{process=\"faucet\"} 8302
 test_varz_value process_state{process=\"gauge\"} 8080 forch
 test_varz_value process_state{process=\"forch\"} 8080 forch
 test_varz_value process_state{process=\"sleep\"} 8080 forch
-test_varz_value process_state{process=\"sleep\"} 8080 random_string
+test_varz_value process_state{process=\"sleep\"} 8080 jkdhfgjksdhg
```
grafnu, line 86: Now that's what I'm talkin' about!

PR #90, opened 2020-04-28
Expose learned macs via Forch varz

didovesei:
Expose learned macs with swith, port, ip, VLAN and ACLs via Forch varz.

Also shared a doc with output examples.


forch/faucet_state_collector.py
```diff
@@ -635,19 +636,19 @@ def _fill_acls_behavior(self, switch_name, acls_map_list, acls_config,
             rules_map_list = acl_map.setdefault('rules', [])
 
             for rule_config in acl_config.rules:
-                cookie = str(rule_config.get('cookie'))
+                num_cookie = rule_config.get('cookie')
```
grafnu, line 639: cookie_num --- num_cookie would be a count of the number of cookies.

forch/forch_metrics.py
```diff
@@ -80,6 +80,10 @@ def _add_vars(self):
                       'No. of RADIUS query responses received from server', Counter)
         self._add_var('process_state', 'Current process state', Gauge, labels=['process'])
 
+        learned_mac_help_text = 'MAC address stored as 64bit number to switch anme, port, VLAN, '\
```
grafnu, line 83: don't use line continuation, this string seems long to have here -- where does it show up?

forch/faucet_state_collector.py
didovesei: Done

forch/forch_metrics.py
didovesei: This is the HELP message that will be shown when scraping the metrics. Updated the message to better describe the new format.

PR #92, opened 2020-04-28
Make source port for RADIUS queries configurable

anurag6:


proto/forch_configuration.proto
```diff
@@ -107,6 +107,9 @@ message OrchestrationConfig {
 
     // Shell command to get RADIUS secret. Can be as simple as 'echo <secret>'
     string radius_secret_helper = 3;
+
+    // Radius port to listen on for responses
```
grafnu, line 111: "Port to listen on for radius responses"

forch/authenticator.py
```diff
@@ -133,6 +134,8 @@ def parse_args(raw_args):
                         help='RADIUS server ip')
     parser.add_argument('-p', '--server-port', type=int, default=1812,
                         help='Server port that freeradius is listening on')
+    parser.add_argument('-l', '--listen-port', type=int, default=0,
```
grafnu, line 137: How is this different than the previous line?  Server ports are typically listen ports?  Is the term "freeradius" in the previous one wrong?  Shouldn't that be "Remote radius server port"?

anurag6: Freeradius is the process that radius servers run. Should mean the same as remote RADIUS server but I can change it.

grafnu:
Let's change it since it's not relevant to this code... it's pulling in an
external fact (that all external radius servers are freeradius) that isn't
relevant at this point.  Doesn't really matter (it's just a comment) but
good to have clear thinking about the various abstractions.

proto/forch_configuration.proto
anurag6: Addressed

anurag6:
Addressed issues. PTAL @grafnu

PR #94, opened 2020-04-29
Add forch config to sys_config API

didovesei:


grafnu:
Is it easy to add faucet_structural in while you're at it?

didovesei:
> Is it easy to add faucet_structural in while you're at it?

Yes, added structural config

PR #96, opened 2020-05-04
Log RADIUS retries in debug.

anurag6:


forch/simple_auth_state_machine.py
```diff
@@ -95,6 +95,9 @@ def handle_sm_timer(self):
         """Handle timer timeout and check.trigger timeout behavior of states"""
         with self._transition_lock:
             if time.time() > self._current_timeout:
+                if self._retry_backoff:
+                    LOGGER.debug('Retrying RADIUS request for src_mac %s. Retry #%s',
+                                 self.src_mac, self._retry_backoff)
```
grafnu, line 100: so _retry_backoff isn't a complete count, since it's capped at the max value.  Maybe remove the cap at the increment and just use min() at the time you calculate the actual backoff time.

anurag6: Makes sense. Changed.

PR #100, opened 2020-05-07
Update faucet configurations used in tests

didovesei:


didovesei:
Added checking for faucet_dp_mac:
(1) Each DP is expected to have faucet_dp_mac
(2) All DPs should have the same faucet_dp_mac

PR #105, opened 2020-05-12
Monitor CPU percent of target processes

didovesei:


forch/local_state_collector.py
```diff
@@ -149,40 +149,49 @@ def _extract_process_state(self, proc_name, proc_count, proc_list):
...
+        except Exception as e:
+            return None, f'Error in extracting info for process {proc_name}: {e}'
 
-        if error:
-            return None, error
+        target_proc_config = self._target_procs[proc_name]
+        if target_proc_config.cpu_percent_threshold:
+            if proc_map['cpu_percent'] > target_proc_config.cpu_percent_threshold:
+                error = f'CPU percent of process {proc_name} is {proc_map["cpu_percent"]}, '
+                error += f'exceeding threashold {target_proc_config.cpu_percent_threshold}'
```
grafnu, line 161: Is this done periodically or on demand from the API?

Also, does it log anything in the case where it happens, or only return an error?  The logging would likely be very useful since it would then correlate in the log file with any triggering activity.

didovesei: This is periodically, along with the existing process polling.

Added a logging at 'error' level.

forch/local_state_collector.py
```diff
@@ -158,7 +158,8 @@ def _extract_process_state(self, proc_name, proc_count, proc_list):
         if target_proc_config.cpu_percent_threshold:
             if proc_map['cpu_percent'] > target_proc_config.cpu_percent_threshold:
                 error = f'CPU percent of process {proc_name} is {proc_map["cpu_percent"]}, '
-                error += f'exceeding threashold {target_proc_config.cpu_percent_threshold}'
+                error += f'exceeding threshold {target_proc_config.cpu_percent_threshold}'
+                LOGGER.error(error)
```
grafnu, line 162: really more of a warning... High CPU in itself is not necessarily an error.

PR #107, opened 2020-05-13
Decouple device behavior from device placement in Faucetizer

didovesei:
In the previous implementation, the static behavior will be used only if both static placement and static behavior for a device is provided.

There are actually 4 posssible combinations:
(static or dynmaic placement) + (static or dynamic behavior)

Changing the implementation such that any of the 4 combinations will be accepted by Faucetizer. Also shared a doc explaining the 4 cases.


forch/faucetizer.py
```diff
@@ -131,31 +135,38 @@ def _faucetize(self):
...
-                if device.behavior.role:
-                    port_cfg['acls_in'] = [f'role_{device.behavior.role}', 'tail_acl']
-                else:
-                    port_cfg['acls_in'] = ['tail_acl']
+                if self._config.unauthenticated_vlan:
+                    port_map['native_vlan'] = self._config.unauthenticated_vlan
+
+        device_placements = {**self._dynamic_devices.device_mac_placements,
+                             **self._static_devices.device_mac_placements}
+        device_behaviors = {**self._dynamic_devices.device_mac_behaviors,
```
grafnu, line 144: what is the intended behavior here if something has been used both as a static device and dynamic device?  How is that clearly disambiguated?

forch/faucetizer.py
```diff
@@ -131,31 +135,38 @@ def _faucetize(self):
...
-                    LOGGER.warning('Device segment does not have a matching vlan: %s, %s',
-                                   mac, device.behavior.segment)
-                    continue
-
-                port_cfg['native_vlan'] = vid
-                if device.behavior.role:
-                    port_cfg['acls_in'] = [f'role_{device.behavior.role}', 'tail_acl']
-                else:
-                    port_cfg['acls_in'] = ['tail_acl']
+                if self._config.unauthenticated_vlan:
```
grafnu, line 139: We need to make sure we're starting to add proper unit testing to changes like this -- esp. when they have algorithmic behavior behind them (which this does).  Until we get the unit test framework in place, can you write some simple test functions that just check some of the cases (e.g. both static & dynamic placements work together properly) and then call them at the end in the main section?

forch/faucetizer.py
didovesei: The intended behavior is that each static placement will overwrite the corresponding dynamic placement, if there is both static and dynamic placement for a device.

"device_placements" here is like a union of the two set of placments. In case of common keys, the static placment will replace the dynamic one.

Should I explicitly list select which info (dynamic/static) to use in a way like if..else..?

grafnu: I think a one-line comment would suffice to describe the intended behavior. Otherwise, e.g., somebody might switch the order b/c "the order of sets doesn't matter". Also then, this should be covered by the unit tests to make sure that the intended behavior is happening.

forch/faucetizer.py
didovesei: Reverted this change. I think it is better to make this behavior change in another PR...

forch/faucetizer.py
didovesei: Added comment. Also added a new test case to cover this "static info overwriting dynamic info" behavior.

forch/faucetizer.py
```diff
@@ -131,31 +135,38 @@ def _faucetize(self):
...
-                    LOGGER.warning('Switch or port not defined in faucet config: %s, %s',
-                                   device.placement.switch, device.placement.port)
-                    continue
 
-                vid = self._segments_to_vlans.get(device.behavior.segment)
-                if not vid:
-                    LOGGER.warning('Device segment does not have a matching vlan: %s, %s',
-                                   mac, device.behavior.segment)
-                    continue
+                port_map['native_vlan'] = self._config.unauthenticated_vlan
```
grafnu, line 139: Can you extract the logic above for if it's an acceptable port above to an appropriately named function?  Would like to know what it's calculating.  Also, not sure that logic is sound e.g. not everything will use lacp.  How about also checking for tagged_vlans?

testing/unit/test_faucetizer.py
```diff
@@ -89,16 +169,137 @@ def tearDown(self):
...
+            ('02:00:00:00:00:03', {'switch': 't2sw2', 'port': 1, 'connected': True}, False),
+            # devices expired
+            ('02:00:00:00:00:01', {'switch': 't2sw2', 'port': 2, 'connected': False}, False),
+            ('02:00:00:00:00:03', {'switch': 't2sw2', 'port': 1, 'connected': False}, False)
+        ]
 
-        self.assertEqual(faucetizer_behavioral_config, expected_behavioral_config)
+        behaviors = [
+            # mocking static behaviors
+            ('02:00:00:00:00:01', {'segment': 'SEG_A', 'role': 'red'}, True),
```
grafnu, line 190: Is there a (negative) test somewhere for a segment that doesn't exist?

testing/unit/test_faucetizer.py
```diff
@@ -89,16 +169,137 @@ def tearDown(self):
...
+                stack: {dp: t1sw1, port: 7}
+        include: []
+        """
+        self._verify_behavioral_config(yaml.safe_load(expected_behavioral_config_str))
+
+        # device expired
+        self._process_device_placement(placements[4])
+        self._process_device_placement(placements[5])
+
+        expected_behavioral_config_str = """
```
grafnu, line 259: can you maybe just modify the expected_behavior_config dict (result from yaml.safe_load) from the previous bit with the explicit expected changes?  Having the complete representation of the structure in both places is tricky to read because it's hard to tell what's different!

testing/unit/test_faucetizer.py
```diff
@@ -89,16 +169,137 @@ def tearDown(self):
...
+        self._process_device_behavior(behaviors[0])
+        self._process_device_behavior(behaviors[1])
+
+        # process dynamic device info
+        self._process_device_behavior(behaviors[2])
+        self._process_device_placement(placements[2])
+        self._process_device_placement(placements[3])
+        self._process_device_behavior(behaviors[3])
+
+        expected_behavioral_config_str = """
```
grafnu, line 209: move definition outside of function -- it's too disruptive to the programmatic flow.  Should be a class variable.

testing/unit/test_faucetizer.py
didovesei: Got it. Added a class variable with basic behavioral config, and updating from there.

testing/unit/test_faucetizer.py
didovesei: Done

testing/unit/test_faucetizer.py
didovesei: Just added device '02:00:00:00:00:04' whose segment does not exist.

forch/faucetizer.py
didovesei: Moved to class method '_is_access_port' and added checking for 'tagged_vlans'.

forch/faucetizer.py
```diff
@@ -131,31 +135,38 @@ def _faucetize(self):
...
+        for mac, device_placement in device_placements.items():
+            device_behavior = device_behaviors.get(mac)
+            if not device_behavior:
+                continue
+
+            switch_cfg = behavioral_faucet_config.get('dps', {}).get(device_placement.switch, {})
+            port_cfg = switch_cfg.get('interfaces', {}).get(device_placement.port)
+
+            if not port_cfg:
+                LOGGER.warning('Switch or port not defined in faucet config: %s, %s',
```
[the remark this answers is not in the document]

PR #108, opened 2020-05-14
Add unit test framework and reorganize tests

didovesei:


.travis.yml
```diff
@@ -25,3 +25,4 @@ env:
     - TEST_SET=reconfig
     - TEST_SET=solo SKIP_STACK=true
     - TEST_SET=stack
+    - TEST_SET=unit SKIP_INSTALL=true SKIP_STACK=true
```
grafnu, line 28: I think we should consider combining this with topology (and possibly renaming to 'base'). The problem is that each high-level entry is fairly expensive in terms of startup-costs, and so adding lots of high-level ones isn't always the the right level of granularity. 

test_base could just call test_topology and test_unit  -- so the underlying modularity is preserved.

bin/test_unit
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -e
```
grafnu, line 1: I think it's time we isolated all of the testing stuff in it's own directory.  Take a look at daq and we should roughly mimic daq/testing -- and this script (along with other test_ scripts) should go there.

testing/unit/test_faucetizer.py
```diff
@@ -0,0 +1,68 @@
...
+import unittest
+
+import forch.faucetizer
+from forch.faucetizer import Faucetizer
+
+TEST_DATA_DIR = os.getenv('TEST_DATA_DIR')
+
+TEST_METHOD_CONFIGS = {
+    'test_faucetize_simple': {
+        'forch_config_file': 'forch_dva.yaml',
```
grafnu, line 13: I think these can just be class variables on the Test Faucetizer class.  It makes it much easier to both reference (don't need to muck about with dictionary lookup), and also override in sub-classes for specific test case setup.  The model is that additional test groups can be created by creating sub-classes of the base class, so we should use OOP design to manage things as much as possible.   See the use of CONFIG in faucet: https://github.com/faucetsdn/faucet/blob/master/tests/unit/faucet/test_valve_stack.py#L43

testing/unit/test_faucetizer.py
```diff
@@ -0,0 +1,68 @@
...
+        self._output_behavioral_config_file = OUTPUT_BEHAVIROAL_CONFIG_FILE
+
+    def setUp(self):
+        self._configure_faucetizer()
+
+    def tearDown(self):
+        self._faucetizer = None
+        os.remove(self._output_behavioral_config_file)
+
+    def _configure_faucetizer(self):
```
grafnu, line 38: configure --> initialize

bin/test_unit
didovesei: Moved all test_* scripts in travis.yaml to testing

.travis.yml
didovesei: Created test_base amd calling test_topology and test_unit there.

testing/unit/test_faucetizer.py
didovesei: Done

testing/unit/test_faucetizer.py
didovesei: I see.. Created a base class and moved the testing stuff into a new test case class that inherits the base one.

forch/utils.py
```diff
@@ -64,3 +64,8 @@ def proto_json(message):
 def dict_proto(message, proto_func, ignore_unknown_fields=False):
     """Convert a standard dict object to a proto object"""
     return json_format.ParseDict(message, proto_func(), ignore_unknown_fields)
+
+
+def text_proto(message_text, proto_func):
```
grafnu, line 69: str_proto

didovesei: Done

PR #110, opened 2020-05-20
Change initial Faucet config behavior

didovesei:
Make unauthenticated_vlan optional
Add optional tail_acl config and add this ACL to unauthenticated host ports
Add unit test for initial Faucet config behavior

### DIFF
--- a/testing/unit/test_faucetizer.py
+++ b/testing/unit/test_faucetizer.py
@@ -54,10 +54,11 @@ class FaucetizerTestBase(unittest.TestCase):
             behavior_tuple[0], dict_proto(behavior_tuple[1], DeviceBehavior),
             behavior_tuple[2])
 
-    def _update_port_config(self, behavioral_config, switch, port, vlan, role):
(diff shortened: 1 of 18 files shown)


testing/unit/test_faucetizer.py
```diff
@@ -58,7 +58,8 @@ def _update_port_config(self, behavioral_config, switch, port, vlan, role=None,
         port_config = behavioral_config['dps'][switch]['interfaces'][port]
         port_config['native_vlan'] = vlan
         port_config['acls_in'] = [f'role_{role}'] if role else []
-        port_config['acls_in'].append(tail_acl)
+        if tail_acl:
```
grafnu, line 61: is this functionally same as the line above?  append X if X is defined?   If so, can you use the same syntax in both places?

didovesei: I think the previous logic is wrong since role-based ACL is independent of tail ACL. Changed the logic a little bit.. PTAL..

PR #111, opened 2020-05-21
Expose DVA state in NOAH switch_state API

didovesei:


forch/faucetizer.py
```diff
@@ -128,6 +130,10 @@ def _is_access_port(self, port_cfg):
             property for property in non_access_port_properties if property in port_cfg]
         return len(port_properties) == 0
 
+    def _update_vlan_state(self, switch, port, state):
+        self._vlan_states.setdefault(switch, {})[port] = state
+
+    # pylint: disable=too-many-branches
```
grafnu, line 136: can you please clean up the function?  Starting to get way too big.

didovesei: Done.. Doing the initialization in another function now. PTAL..

PR #112, opened 2020-05-21
Expose RADIUS result info in NOAH

anurag6:


etc/test_access.out
```diff
@@ -1,4 +1,4 @@
-testing/test_access
+./testing/test_access
```
grafnu, line 1: remove. leading ./ is redundant

PR #115, opened 2020-05-27
Raise exception if no access switch found

anurag6:


forch/faucet_state_collector.py
```diff
@@ -1114,6 +1114,8 @@ def _update_stack_links_stats(self, timestamp):
     def _update_learned_macs_metrics(self):
         for mac in self.learned_macs:
             switch, port = self._get_access_switch(mac)
+            if not (switch and port):
```
grafnu, line 1117: should be an else clause of the following if statement

anurag6: Addressed.

anurag6:
> assert switch and port, 'Could not find access switch for mac %s' % mac
> self._update_learned_macs_metric(mac, switch, port)

Makes sense

PR #118, opened 2020-06-02
Add ACLs counting topo and corresponding tests

didovesei:


.travis.yml
```diff
@@ -24,4 +24,5 @@ env:
     - TEST_SET=base SKIP_INSTALL=true SKIP_STACK=true
     - TEST_SET=reconfig
     - TEST_SET=solo SKIP_STACK=true
+    - TEST_SET=counting SKIP_STACK=true
```
grafnu, line 27: This should not be its own top-level container -- they are too expensive to spin up for each little thing, it takes a good 10 minutes to just setup the build.  Likely should be combined with test_solo -- see testing/test_base for an example of what that might look like.

topo/counting/faucet.yaml
```diff
@@ -0,0 +1,181 @@
...
+    faucet_dp_mac: 0e:00:00:00:02:01
+    hardware: Generic
+    lacp_timeout: 5
+    interfaces:
+      1: {description: IoT Device, native_vlan: 100, max_hosts: 1}
+      2: {description: IoT Device, native_vlan: 100, max_hosts: 1}
+      3: {description: IoT Device, native_vlan: 100, max_hosts: 1}
+      4: {description: IoT Device, native_vlan: 100, max_hosts: 1}
+      5: {description: IoT Device, native_vlan: 100, max_hosts: 1}
+      6: {description: IoT Device, native_vlan: 100, max_hosts: 1}
```
grafnu, line 57: are all of these necessary for your test?  If not I'd trim them down because all they're doing then is making life more complicated for us!

.travis.yml
didovesei: Right, merged two tests into testing/test_topos.

topo/counting/faucet.yaml
didovesei: Removed all the unnecessary host ports.

PR #119, opened 2020-06-04
Check if port change evnet's port exists in dp config

didovesei:


forch/faucet_state_collector.py
```diff
@@ -871,6 +871,12 @@ def get_host_path(self, src_mac, dst_mac, to_egress):
     def process_port_state(self, timestamp, name, port, state):
         """process port state event"""
         with self.lock:
+            switch_config = self.faucet_config.get(name)
+            if not switch_config:
```
grafnu, line 875: for things like this assert is preferred:
assert switch_config, 'Switch %s is not in faucet config' % name

didovesei: Updated. PTAL..

PR #123, opened 2020-06-12
Add Forch varz for packet rate states for vlans

didovesei:


forch/faucet_state_collector.py
```diff
@@ -181,15 +181,18 @@ def _update_packet_count_states(self, vlan_counts, interval):
...
                     vlan_id, threshold, rate)
-                last_vlan_map[PACKET_RATE_STATE] = State.broken
+                rate_state = State.broken
             else:
-                last_vlan_map[PACKET_RATE_STATE] = State.healthy
+                rate_state = State.healthy
+
+            last_vlan_map[PACKET_RATE_STATE] = rate_state
+            self._forch_metrics.update_var(
+                'vlan_packet_rate_state', rate_state, labels=[vlan_id])
```
grafnu, line 195: should expose actual counts (could also do state too, but we want the raw numbers exposed because they can be used to make pretty graphs)

didovesei: Expose the raw counts now. PTAL..

PR #125, opened 2020-06-17
Augment ACLs in faucet.yaml and verify role-based ACLs existence

didovesei:
* Add cookies to the ACLs that are defined in structural config file
* Do not assign role-based ACLs to port if no ACL defined for this role

### DIFF
--- a/forch/faucetizer.py
+++ b/forch/faucetizer.py
@@ -104,7 +104,7 @@ class Faucetizer:
 
             tail_acl_name = self._config.tail_acl
             if tail_acl_name and not self._has_acl(tail_acl_name):
-                raise Exception('No ACL is defined for tail ACL %s', tail_acl_name)
+                raise Exception('No ACL is defined for tail ACL %s' % tail_acl_name)
 
             self._behavioral_include = behavioral_include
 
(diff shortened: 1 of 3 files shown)


forch/faucetizer.py
```diff
@@ -178,9 +195,17 @@ def _faucetize(self):
...
             if device_behavior.role:
-                port_cfg['acls_in'] = [f'role_{device_behavior.role}']
+                acl_name = f'role_{device_behavior.role}'
+                if self._has_acl(acl_name):
+                    port_cfg['acls_in'] = [acl_name]
+                else:
+                    port_cfg['acls_in'] = []
+                    LOGGER.error('No ACL defined for role %s', device_behavior.role)
+            else:
+                port_cfg['acls_in'] = []
```
grafnu, line 208: The two lines below this seem a bit sketchy -- can you re-write?  "setdefault" seems redundant with the code above, and then if tail_acl is NOT defined then it's not called at all, which means that either 1) it's not necessary to call at all, or 2) there's some weird corner-case when it should be called and isn't.  So, I think you can just take it out and directly call .append when appropriate.

didovesei: Right! The setdefault is useless as 'acls_in' is always set when processing the role. PTAL...

PR #128, opened 2020-07-04
Account for port role while calculating egress link state. Change is_egress_port to fix error querying lag state when varz are queried. 

anurag6:


forch/faucet_state_collector.py
```diff
@@ -1350,7 +1368,11 @@ def _get_egress_port(self, switch):
         return None
 
     def _is_egress_port(self, switch, port):
```
anurag6, line 1370: The change fixes the error observed in prod where LAG state returns None. Seems that's the case when lag state is restored form varz.

forch/faucet_state_collector.py
```diff
@@ -310,6 +311,23 @@ def _restore_dataplane_state_from_metrics(self, metrics):
         timestamp = time.time()
         self._update_stack_topo_state(timestamp, link_graph, stack_root, dps)
 
+    def _restore_lag_state_from_metrics(self, metrics):
+        """Restores dataplane state from prometheus metrics. relies on STACK_STATE being restored"""
```
grafnu, line 315: nit: relies --> Relies

PR #133, opened 2020-07-28
Expose faucet config warning through API and Varz

didovesei:
Example of varz:
faucet_config_warning_count 2.0
faucet_config_warning{key="faucet_dp_mac",message="faucet_dp_mac for DPs are not identical"} 1.0
faucet_config_warning{key="nz-kiwi-t1sw2",message="faucet_dp_mac not defined"} 1.0

### DIFF
--- a/forch/forchestrator.py
+++ b/forch/forchestrator.py
@@ -633,7 +633,7 @@ class Forchestrator:
         self._metrics.update_var(
             'faucet_config_warning_count', len(self._config_summary.warnings))
         for warning_key, warning_msg in self._config_summary.warnings.items():
-            self._metrics.update_var('faucet_config_warning', 0, warning_msg)
+            self._metrics.update_var('faucet_config_warning', 0, [warning_key, warning_msg])
 
     def _populate_versions(self, versions):
         versions.forch = __version__
--- a/forch/forchestrator.py
+++ b/forch/forchestrator.py
@@ -351,6 +351,7 @@ class Forchestrator:
 
     def _restore_faucet_config(self, timestamp, config_hash):
         config_info, faucet_dps, _ = self._get_faucet_config()
+        self._update_config_warning_varz()
         assert config_hash == config_info['hashes'], 'config hash info does not match'
         self._faucet_collector.process_dataplane_config_change(timestamp, faucet_dps)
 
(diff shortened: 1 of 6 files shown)


forch/forchestrator.py
```diff
@@ -687,8 +694,12 @@ def get_sys_config(self, path, params):
         """Get overall config from faucet config file"""
         try:
             _, _, behavioral_config = self._get_faucet_config()
+            behavioral_config_map = {
+                'content': behavioral_config,
+                'warnings': dict(self._config_summary.warnings)
+            }
             reply = {
-                'faucet_behavioral': behavioral_config,
+                'faucet_behavioral': behavioral_config_map,
```
grafnu, line 702: This doesn't seem right and isn't logically consistent.  The warnings are not part of the behavioral config. Seems like the top-level should be faucet (parallel to forch) and then 'content' would be renamed to 'behavorial'.

forch/forchestrator.py
```diff
@@ -628,6 +629,12 @@ def _validate_config(self, config):
                 warnings.append(('faucet_dp_mac', 'faucet_dp_mac for DPs are not identical'))
         return warnings
 
+    def _update_config_warning_varz(self):
+        self._metrics.update_var(
+            'faucet_config_warning_count', len(self._config_summary.warnings))
+        for warning_key, warning_msg in self._config_summary.warnings.items():
+            self._metrics.update_var('faucet_config_warning', 0, [warning_key, warning_msg])
```
grafnu, line 636: should be 1 not 0, yes, there's no case where it's anything, but a 0 value implies "this is not true" which is not the case.

testing/test_stack
```diff
@@ -66,6 +66,8 @@ curl http://$CONTROLLER_NAME:9019/protos.html > $OUT_DIR/protos.html
 fgrep "forch/proto/system_state.proto" $OUT_DIR/protos.html || exit 1
 curl http://$CONTROLLER_NAME:9019/sys_config > $OUT_DIR/sys_config.json
 jq '.faucet_behavioral.dps."nz-kiwi-t2sw3"'.faucet_dp_mac $OUT_DIR/sys_config.json | tee -a $TEST_RESULTS
+jq '.faucet_behavioral.warnings.faucet_dp_mac' $OUT_DIR/sys_config.json | tee -a $TEST_RESULTS
+test_varz_value faucet_config_warning_count 8302 forch
```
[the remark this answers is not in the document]
didovesei: yes, just added in etc/test_stack.out

forch/forchestrator.py
didovesei: updated the output structure. PTAL

forch/forchestrator.py
didovesei: Makes sense. Changed to 1

forch/forchestrator.py
didovesei: Now it looks like:

- faucet
   * behavioral
   * structural
   * warnings
- forch
   * ...

PR #134, opened 2020-07-29
Base changes to support OWCA emu

anurag6:


bin/run_faucet
```diff
@@ -36,6 +36,7 @@ fi
 if [ -n "$1" ]; then
     CONTAINER=$CONTAINER-$1
     INSTDIR=$INSTDIR/forch-faucet-$1
+    id=$1
```
grafnu, line 39: something longer than just "id" please, what kind of ID is it?

PR #136, opened 2020-07-30
Add github workflows for integration tests

didovesei:
There could have been a more parameterized workflow definition using matrix with nested parameters. I personally prefer this one since it is more straight forwarded and would be easier to maintain...

### DIFF
--- a/bin/setup_base
+++ b/bin/setup_base
@@ -1,5 +1,8 @@
 #!/bin/bash -e
 
+echo "Getting interface..."
+route -n | grep -E '^0\.0\.0\.0'
+
 ROOT=$(realpath $(dirname $0)/..)
 cd $ROOT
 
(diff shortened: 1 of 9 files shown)


bin/setup_base
```diff
@@ -15,7 +21,7 @@ fi
 
 pip3 install -r etc/requirements.txt
 
-DEF_IFACE=`route -n | fgrep UG | awk '{print $8}'`
+DEF_IFACE=`route -n | grep -E '^0\.0\.0\.0' | awk '{print $8}'`
```
grafnu, line 24: Use 'egrep \sUG\s' -- Haoli had to do this on another PR as well, so just want to have a consistent solution. @henry54809

didovesei: Updated.

PR #139, opened 2020-08-05
Avoid empty acls_in for a port during DVA

didovesei:


testing/unit/test_faucetizer.py
```diff
@@ -319,6 +319,123 @@ def test_devices_learned_and_authenticated(self):
...
 
 
+class FaucetizerBehaviorWithoutTailACLTestCase(FaucetizerTestBase):
+    """Test Faucetizer's behavior with no tail_acl setting in Forch config"""
+
+    ORCH_CONFIG = """
+    unauthenticated_vlan: 100
+    """
+
+    FAUCET_STRUCTURAL_CONFIG = """
```
grafnu, line 329: this seems very basic (the structural config) -- should it be promoted to a superclass to prevent repetition?  Is there anything unique to this test case about it?

testing/unit/test_faucetizer.py
```diff
@@ -319,6 +319,123 @@ def test_devices_learned_and_authenticated(self):
...
+                stack: {dp: t2sw2, port: 7}
+              23:
+                lacp: 3
+          t2sw1:
+            dp_id: 121
+            interfaces:
+              1:
+                description: HOST
+                max_hosts: 1
+                native_vlan: 100
```
grafnu, line 374: why is 100 here?  Likewise, can this generic config template be promoted to the superclass?

forch/faucetizer.py
```diff
@@ -208,12 +208,9 @@ def _faucetize(self):
                 if self._has_acl(acl_name):
                     port_cfg['acls_in'] = [acl_name]
                 else:
-                    port_cfg['acls_in'] = []
                     LOGGER.error('No ACL defined for role %s', device_behavior.role)
-            else:
-                port_cfg['acls_in'] = []
             if self._config.tail_acl:
-                port_cfg['acls_in'].append(self._config.tail_acl)
+                port_cfg.setdefault('acls_in', []).append(self._config.tail_acl)
```
grafnu, line 213: I think maybe your title is misleading?  Seems like this will configure the acls_in with the tail_acl even if the device does not have a role?

grafnu, line 213: try to be consistent in the code in the treatment of this setting, so, e.g., above you should use setdefault(() as well rather than setting = [acl_name] -- the reason is consistency and more straightforward flow so I don't need to worry abotut "has the variable been previously initialized".

didovesei: Updated

didovesei: I think the problem is that in line 178, faucetizer already added tail_acl to all the host ports. If we use setdefault in line 209, it will append the role acl to the existing list ([tail_acl]), which will finally result in [tail_acl, role_acl, tail_acl].  I think maybe it is correct to explicitly overwrite the port config in line 209, since we have a device behavior for that port and hence that port needs to reconfigured based on the device behavior.

grafnu: the right solution is to add the tail_acl at the end of the flow, not the beginning... overwriting likely isn't the way to go.

didovesei:
> why would tail_acl already be there? This should be very simple logic, so it's a red flag that there's corner cases...

It is because we apply the tail acl in the host ports initialization (self._initialize_host_ports()). Moved that part out of the initialization. Now it applies tail acl after all device behaviors are processed. PTAL..

grafnu:
Looks good... approval still stands!

PR #143, opened 2020-08-07
Enable Faucetizer sequester devices in testing VLANs

didovesei:
With this change, Faucetizer expects a state machine of the device states exists outside Faucetizer. For a device that needs to be sequestered, the state machine will call process_device_behavior() with "testing" segment. For a device that can be operational, it will call process_device behavior() with the authenticated segment.


forch/faucetizer.py
```diff
@@ -141,11 +144,24 @@ def _augment_include_file_name(self, file_name):
         base_file_name, ext = os.path.splitext(file_name)
         return base_file_name + INCLUDE_FILE_SUFFIX + ext
 
-    def _is_access_port(self, port_cfg):
+    def _get_port_type(self, port_cfg):
+        testing_port_identifier = (self._config.orch_testing_config.testing_port_identifier or
+                                   TESTING_PORT_IDENTIFIER_DEFAULT)
+        if testing_port_identifier in port_cfg.get('description'):
+            return 'testing'
```
grafnu, line 151: create named constants for these (e.g. PORT_TYPE.TESTING)

forch/faucetizer.py
```diff
@@ -141,11 +144,24 @@ def _augment_include_file_name(self, file_name):
...
         non_access_port_properties = ['stack', 'lacp', 'output_only', 'tagged_vlans']
         port_properties = [
             property for property in non_access_port_properties if property in port_cfg]
-        return len(port_properties) == 0
+        return 'access' if len(port_properties) == 0 else 'other'
+
+    def _calcuate_available_tesing_vlans(self):
+        if not self._config.orch_testing_config.testing_vlans:
+            return
+        starting_vlan = self._config.orch_testing_config.testing_vlans[0]
```
grafnu, line 160: somewhere there needs to be a check that the stated vlan range does not overlap with any potential assigned segment vlans.

proto/forch_configuration.proto
```diff
@@ -88,6 +88,9 @@ message OrchestrationConfig {
   // authentication configurations
   AuthConfig auth_config = 7;
 
+  // orchestrated testing configurations
+  OrchestratedTestingConfig orch_testing_config = 11;
```
grafnu, line 92: I think just fot_config  -- trying to propagate orch_testing everywhere will be very verbose.  Normally I don't like abbreviations, but I think in this case it works better.  We're starting to call it "Faucet Orchestrated Testing" just for this reason (because OT is a bad abbreviation, FOT is better).

proto/forch_configuration.proto
```diff
@@ -123,6 +126,18 @@ message OrchestrationConfig {
     // Port to listen on for RADIUS responses
     int32 source_port = 4;
   }
+
+  // encapsulating orchestrated testing configurations
+  message OrchestratedTestingConfig {
```
grafnu, line 131: this would then be FotTestingConfig

proto/forch_configuration.proto
```diff
@@ -123,6 +126,18 @@ message OrchestrationConfig {
...
+    // testing segment. If defined, automatic sequestering is enabled
+    string testing_segment = 11;
+
+    // range of testing VLANs, represented in a list of starting and ending VLAN
```
grafnu, line 135: is it a range or a list?  If it's a range then there should be two specific values. If it's an arbitrary list then it would be repeated.

forch/faucetizer.py
didovesei: Added PortType enum proto in shared_constants.proto

forch/faucetizer.py
didovesei: Added the check in this function.

proto/forch_configuration.proto
didovesei: updated

proto/forch_configuration.proto
didovesei: Done

proto/forch_configuration.proto
didovesei: Separated it into two values.

didovesei:
> Nice encapsulation in the faucetizer class -- that's exactly the right way to do it (state machine external).
> 
> Need to add some (unit?) tests for this. Note there is a new "test_ot" system that Anurag put in place that has a python test running... Ideally would use that.

Added the test inside the ot tests.

forch/faucetizer.py
didovesei: Acutally moved the check into the initialization of Faucetizer...

forch/faucetizer.py
```diff
@@ -144,24 +148,38 @@ def _augment_include_file_name(self, file_name):
...
                                    TESTING_PORT_IDENTIFIER_DEFAULT)
-        if testing_port_identifier in port_cfg.get('description'):
-            return 'testing'
+        if testing_port_identifier in port_cfg.get('description', ""):
+            return PortType.testing
         non_access_port_properties = ['stack', 'lacp', 'output_only', 'tagged_vlans']
         port_properties = [
             property for property in non_access_port_properties if property in port_cfg]
-        return 'access' if len(port_properties) == 0 else 'other'
+        return PortType.access if len(port_properties) == 0 else PortType.unknown
```
grafnu, line 177: 'unknown' is not the same as 'other' -- in this case, I think you do want 'other' (since it's not testing or access), while 'unknown' would be 0 or None (unassigned). So, the proto should have both, and this use should be 'other'.

proto/forch_configuration.proto
```diff
@@ -89,7 +89,7 @@ message OrchestrationConfig {
   AuthConfig auth_config = 7;
 
   // orchestrated testing configurations
-  OrchestratedTestingConfig orch_testing_config = 11;
+  FOTConfig fot_config = 11;
```
grafnu, line 92: This should be FotConfig -- the reason is that in this context 'fot' is a word, so it maps cleanly to fot_config -- otherwise, following the rules of conversion FOTConfig would map to f_o_t_config

topo/bond/faucet/faucet.yaml
```diff
@@ -8,7 +8,7 @@ dps:
       priority: 1
     interfaces:
       4:
-        description: trunk
+        description: TESTING trunk
```
grafnu, line 11: where/how is this used?   The test you have is a unit test so don't use this bit.  What effect does putting TESTING here have and how does it manifest itself?

forch/faucetizer.py

proto/forch_configuration.proto
didovesei: Done

topo/bond/faucet/faucet.yaml
didovesei: My idea was to use it in the test where the Forch "testing_segment" was set to TESTING. But you are right it was not useful... I reverted the change and set "testing_segment" to "trunk".

testing/integration/test_ot.py
```diff
@@ -26,15 +32,46 @@ def test_ot_sequester(self):
...
-        interface['native_vlan'] = 272
-        self._write_faucet_config(config)
+        self._generate_sequestering_config()
         time.sleep(5)
+
         self.assertTrue(self._ping_host('forch-faux-1', '192.168.2.1'))
         self.assertFalse(self._ping_host('forch-faux-1', '192.168.1.2'))
         self._clean_stack()
 
+    def _generate_sequestering_config(self):
```
grafnu, line 42: Is this actually an integration test or a unit test?  If it's a unit test, then it should be in a separate class FotUnitTests(UnitTestBase)  -- compared with class FotIntegrationTests(IntegrationTestBase):

didovesei: Added unit_base.py in python_lib and concrete unit test case in test_fot.py.

PR #147, opened 2020-08-08
Refactor python test runner

grafnu:


testing/python_test
```diff
@@ -0,0 +1,46 @@
...
+
+if [ $# != 1 ]; then
+    echo $0 [test_name]
+    false
+fi
+
+TEST_NAME=$1
+shift
+
+TEST_DIR=testing/python_lib
```
anurag6, line 24: Should we just rename it to python_test since we have the test suite files themselves here?

grafnu: There's already a python_test in that directory (this file itself), so can't use that name...

PR #153, opened 2020-08-13
Auto reload segemnts, static placements and static behaviors files

didovesei:


testing/python_lib/test_faucetizer.py
```diff
@@ -26,25 +26,31 @@ def __init__(self, *args, **kwargs):
...
 
     def _initialize_faucetizer(self):
         self._orch_config = str_proto(self.ORCH_CONFIG, OrchestrationConfig)
 
         self._faucetizer = Faucetizer(
-            self._orch_config, self._temp_structural_config_file, self.SEGMENTS_TO_VLANS,
+            self._orch_config, self._temp_structural_config_file,
             self._temp_behavioral_config_file)
         self._faucetizer.reload_structural_config()
+        self._faucetizer.reload_segments_to_vlans(self._temp_segments_vlans_file)
```
grafnu, line 53: Does anything in here actually check that the file was automatically reloaded when it changes on disk, or is it only testing that it can load the file from disk?  I would expect some sequence like: setup initial file, check results, modify file, check results have changed.

didovesei: This is more like a unit test that purely test faucetizer itself. The file reload is managed by the file watcher. I think I can add it to the DVA integration test.

PR #158, opened 2020-08-18
Fix test_access in Forch test suite

didovesei:


didovesei:
After removed vlan 100 from egress port, Guage is not creating the "flow_packet_count_vlan" anymore for vlan 100. Removing vlan 100 related checking from the test.

PR #159, opened 2020-08-20
Add gRPC interface to receive device testing results

didovesei:


proto/grpc/device_testing.proto
```diff
@@ -0,0 +1,22 @@
...
+ * gRPC service to receive testing results
+ */
+service DeviceTesting {
+    rpc ReportTestingResult (DeviceTestingResult) returns (Empty) {}
+}
+
+/*
+ * Device testing result
+ */
+message DeviceTestingResult {
```
grafnu, line 19: I think this should be DeviceTestingState, not result. e.g., what happens if there's a set of predefined states in a file?  Exact same structure, just not a result anymore.

proto/grpc/device_testing.proto
```diff
@@ -0,0 +1,22 @@
...
+    string mac = 1;
+    bool passed = 2;
```
grafnu, line 21: make this an enum not a bool. Initial entries can be { unknown, testing, passed } -- it will just give us more room in the future to expand to include other states -- e.g. when an anomaly is detected or the device is explicitly cleared.

proto/grpc/device_testing.proto
didovesei: Right, changed to DeviceTestingState.

didovesei: Maybe DeviceTestingState is not a good name... Will udpate..

proto/grpc/device_testing.proto
didovesei: Added TestingState enumus.

forch/device_testing_server.py
```diff
@@ -1,13 +1,14 @@
...
-import grpc
 import logging
 
+import grpc
+
 import forch.proto.grpc.device_testing_pb2_grpc as device_testing_pb2_grpc
 from forch.proto.shared_constants_pb2 import Empty
 
-LOGGER = logging.getLogger('cstate')
+LOGGER = logging.getLogger('grpcserver')
```
grafnu, line 11: dtservicer?  This is more specific to device-testing than it is to grpc.  What happens if there's another grpc RPC somewhere?

didovesei: Updated.

PR #160, opened 2020-08-21
Add loop detection logic and test

anurag6:


anurag6:
Refactored code and added a positive test case. Merging in the morning if you find no issues with it. @grafnu

PR #162, opened 2020-08-22
Add (failing) test for orphaned switch

grafnu:


grafnu:
Pending proper integration with Faucet 1.9.49 -- which has the new stacking algo but breaks some of our tests.

anurag6 (approved):

PR #163, opened 2020-08-24
test_stack tracks delta change_counts

grafnu:


bin/stack_functions
```diff
@@ -274,6 +274,21 @@ function restart_forch {
 
     # Wait for system to warm-up, including background process/cpn checks.
     sleep 20
+
+    unset change_counts
+    declare -g -A change_counts
+}
+
+# Do a delta-query fetch diff for change count totals
+function dq {
```
anurag6, line 283: I would point out abbreviating but it really goes well with jq in the lines below.

grafnu:
Yah -- one of those cases where sometimes rules are meant to be broken, I
think!

PR #164, opened 2020-08-27
Forch inside faucet container

anurag6:


.dockerignore
```diff
@@ -1,4 +1,3 @@
 inst/
 dist/
 build/
-.git/
```
anurag6, line 4: I wanted to copy the .git folder of forch so as to make it easier to install forch in the container. pbr needs git tags.

bin/run_faucet
```diff
@@ -17,6 +17,7 @@ docker_volumes=
 port_opts=
 local=
 faucet_id=
+fot=
```
grafnu, line 20: I think this needs to be a new container ("controller") not an extension of faucet.  You can start with the faucet one and add more stuff to it, but it's not "faucet" anymore.  Inside of this controller container would be { faucet, forch, gauge, daq, XXX } -- rather than (e.g. what you have now) one container that's { faucet, forch} and another one that's { gauge }. Basically, the new "controller" container should be an encapsulation of what is running on the controller.

.dockerignore
grafnu: Not a good idea to do it this way.  The .git directory can become arbitrarily large so can make things get crazy slow -- and esp doing this as a global wildcard.  If you need the git tags inside of the repo, you should do a clean checkout inside of the build container, e.g. https://github.com/grafnu/daq/blob/master/docker/modules/Dockerfile.test_bacnet#L10

bin/setup_stack
```diff
@@ -105,7 +111,7 @@ for index in 1; do
     cp -r topo/$mode/forch inst/$controller/
     cp -r topo/$mode/faucet inst/$controller/
     docker kill $controller || true
-    bin/run_faucet $local $index $((6000 + index)) $((8000 + index))
+    bin/run_faucet $local $fot $index $((6000 + index)) $((8000 + index))
```
grafnu, line 114: as per other comment, rather than $fot here as an option to run_faucet, it should look something like:
```
if [ -n "$fot" ]; then
  bin/run_controller $local $index ... (whatever options)
else
  ... (the stuff that's currently done with bin/run_faucet)
fi
```

etc/Dockerfile.faucet
```diff
@@ -2,11 +2,13 @@
 
 FROM faucet/python3:5.0.1
 
-RUN apk add -q tcpdump iptables sudo
+RUN apk add -q tcpdump iptables sudo linux-headers
```
grafnu, line 5: so, as per other comments, this should be a copy of Dockerfile.faucet into Dockerfile.controller then looks more or less exactly like this file.

.dockerignore
anurag6: Makes sense. Though I wanted to move current codebase instead of git checkout so that we could test local changes that way. An alternative is to simply skip .git and do a git init and then let it carry the 0.0.0 version internally. If the version really matters, I could pass it as an argument some how. Do you thin the version inside the container is important?

bin/run_faucet
anurag6: That sounds liek a lot of code duplication though. An alternative could be to rename current scripts to controller and let the flag decide whether to add daq and forch into it.

etc/Dockerfile.faucet
anurag6: Same as before. Should I just rename the whole thing to controller so that it makes more sense semantically.

.dockerignore
anurag6: Resolved.

bin/run_faucet
anurag6: Separation done.

bin/setup_stack
anurag6: Done.

etc/Dockerfile.faucet
anurag6: Separated.

bin/build_docker
```diff
@@ -82,7 +85,7 @@ tail build/docker_*.log
 
 docker images | fgrep forch/ | fgrep latest
 images=$(docker images | fgrep forch/ | fgrep latest | wc -l)
-if [ $images != 3 ]; then
+if [ $images != 4 ]; then
```
grafnu, line 88: fix next line

bin/setup_stack
```diff
@@ -100,12 +106,20 @@ sudo modprobe bonding
...
     mkdir -p inst/$controller/
     cp -r topo/$mode/forch inst/$controller/
     cp -r topo/$mode/faucet inst/$controller/
     docker kill $controller || true
-    bin/run_faucet $local $index $((6000 + index)) $((8000 + index))
+    if [ -n "$fot" ]; then
+        bin/run_controller $local $index $((6000 + index)) $((8000 + index))
+    else
+        bin/run_faucet $local $index $((6000 + index)) $((8000 + index))
+    fi
```
grafnu, line 122: gauge should be with run_faucet, so it's either { controller } or  { faucet, gauge } not { controller, gauge } and { faucet, gauge }

bin/install-forch.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/ash
+
+set -euo pipefail
```
grafnu, line 3: install_forch.sh  (underscore not dash in filename)

bin/build_docker
anurag6: Done

bin/setup_stack
anurag6: Added gauge to container

bin/install-forch.sh
anurag6: renamed file

bin/run_controller
```diff
@@ -0,0 +1,170 @@
...
+fi
+
+if [ -n "$1" ]; then
+    CONTAINER=$CONTAINER-$1
+    INSTDIR=$INSTDIR/forch-controller-$1
+    faucet_id=$1
+    shift
+fi
+
+if [ -n "$1" ]; then
```
grafnu, line 39: this likely isn't the best way to do this because it ends up as a string of numerical arguments that all look the same... but,  don't need to fix it now -- just realize that it's a messy way to do it. Something that works for n<=3 often doesn't scale for n>3 since it tips over from "simple is easy" to "simple is messy"

bin/run_controller
```diff
@@ -0,0 +1,170 @@
...
+
+forch_port="-p 500$faucet_id:9019"
+docker_envs+="\
+    --env CONTROLLER_NAME=$CONTAINER \
+    --env PROMETHEUS_PORT=9302 \
+    --env FAUCET_EVENT_SOCK=/var/log/faucet/faucet_event.sock \
+    --env FAUCET_CONFIG_DIR=/etc/faucet \
+    --env FORCH_CONFIG_DIR=/etc/faucet \
+    --env WSAPI_LISTEN_HOST=localhost \
+    --env WSAPI_LISTEN_PORT=4000 \
```
grafnu, line 131: likewise, this is starting to get to the point where it will be a lot cleaner to have all this in a file that's then dynamically loaded into the env at runtime.  Again, not something required to fix now but to ponder going forward.

bin/setup_stack
```diff
@@ -100,13 +106,21 @@ sudo modprobe bonding
...
+        controller=forch-faucet-$index
+    fi
     mkdir -p inst/$controller/
     cp -r topo/$mode/forch inst/$controller/
     cp -r topo/$mode/faucet inst/$controller/
     docker kill $controller || true
-    bin/run_faucet $local $index $((6000 + index)) $((8000 + index))
-    bin/run_faucet $local gauge $index $((7000 + index)) $((9000 + index))
+    if [ -n "$fot" ]; then
+        bin/run_controller $local $index $((6000 + index)) $((8000 + index)) $((7000 + index)) $((9000 + index))
```
grafnu, line 119: this is where the endless args get messy!  (again, just FYI)

bin/setup_stack
```diff
@@ -100,13 +106,21 @@ sudo modprobe bonding
 bin/run_faucet $local corp 6683
 
 for index in 1; do
-    controller=forch-faucet-$index
+    if [ -n "$fot" ]; then
+        controller=forch-controller-$index
```
grafnu, line 110: container= not controller=

misc/gauge_go
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -ex
+
+/usr/bin/gauge --ryu-config-file=${GAUGE_RYU_CONF} --ryu-ofp-tcp-listen-port=${GAUGE_LISTEN_PORT} --ryu-wsapi-host=${WSAPI_LISTEN_HOST} --ryu-wsapi-port=${WSAPI_LISTEN_PORT} --ryu-app=ryu.app.ofctl_rest &
```
grafnu, line 3: break onto multiple lines

bin/run_controller
anurag6: I will keep this in mind when I do the test_fot script.

bin/run_controller
anurag6: makes sense.

bin/setup_stack
anurag6: Changed name to container.

misc/gauge_go
anurag6: Broke up command

PR #170, opened 2020-09-16
Connect port state manager to Forch orchestration components

didovesei:


forch/port_state_manager.py
```diff
@@ -72,32 +78,69 @@ def _handle_authenticated_state(self):
...
-    def handle_authenticated_device(self, mac):
-        """initialize or update the state machine for an authenticated device"""
+        isolation_behavior = device_behavior.isolation_behavior
+        if isolation_behavior:
+            self._static_port_behaviors[mac] = isolation_behavior
+
+        if device_behavior.segment:
+            self.handle_auth_result(mac, device_behavior, static=True)
+
+    def handle_auth_result(self, mac, device_behavior, static=False):
```
grafnu, line 109: what about this says it's an auth result?  What happens if the device_behavior comes from something other than auth?  (My point is your function name is overly narrow)

forch/faucetizer.py
```diff
@@ -18,6 +18,10 @@
 
 INCLUDE_FILE_SUFFIX = '_augmented'
 TESTING_PORT_IDENTIFIER_DEFAULT = 'TESTING'
+BEHAVIOR = 'behavior'
```
grafnu, line 21: these are overly generic for top-level constants, so should be qualified somehow... e.g. so it reads DEVICE.STATIC or DEVICE_STATIC

forch/port_state_manager.py
didovesei: Changed to handle_device_behavior()

forch/faucetizer.py
didovesei: Updated. PTAL..

grafnu (approved):

PR #185, opened 2020-09-30
Display error and disable authentication when static behavior file is missing

didovesei:


forch/forchestrator.py
```diff
@@ -230,10 +235,21 @@ def _process_static_device_behavior(self):
     def _reload_static_device_behavior(self, file_path):
         self._port_state_manager.clear_static_device_behaviors()
 
-        devices_state = yaml_proto(file_path, DevicesState)
+        try:
+            devices_state = yaml_proto(file_path, DevicesState)
+        except Exception as error:
+            msg = f'DVA disabled: could not load static behavior file {file_path}'
```
grafnu, line 241: Make sure there is a log message somewhere that indicates when this condition is cleared (the file appears).

grafnu, line 241: Message should be that "Authentication is disabled" -- since it's not just DVA that might be affected by this.

forch/forchestrator.py
```diff
@@ -345,8 +361,10 @@ def _process_device_behavior(self, mac, device_behavior, static=False):
     def handle_auth_result(self, mac, access, segment, role):
         """Method passed as callback to authenticator to forward auth results"""
         self._faucet_collector.update_radius_result(mac, access, segment, role)
-        device_behavior = DeviceBehavior(segment=segment, role=role)
-        self._port_state_manager.handle_device_behavior(mac, device_behavior)
+        with self._lock:
+            if not self._should_ignore_auth_result:
```
grafnu, line 365: add warning log here that authentication ignored for this particular device

grafnu (commented):
Just minor comments about logging messages

forch/forchestrator.py
didovesei: Done. Added logging below

didovesei: Updated

forch/forchestrator.py

grafnu (approved):

PR #186, opened 2020-10-06
Rename protobuf messages used in DVA and FOT

didovesei:


proto/shared_constants.proto
```diff
@@ -140,8 +140,8 @@ message PortType {
 /*
  * Port behaviors
  */
-message PortBehavior {
-  enum Behavior {
+message DeviceEvent {
```
grafnu, line 143: PortBehavior was better, these (as a message construct) aren't "events" -- they may be used as an event, but that's a use of it, not what it is.  PortBehavior, SequestedState, SequesteredBehavior all seem appropriate to me.

DeviceBehavior wouldn't be good because it's not actually the behavior of the device.

PortState is too generic (port up/down?), so also not good here.

grafnu (commented):
Half there... the use of "Event" is problematic in this context.

proto/shared_constants.proto
didovesei: Got it, changed back to PortBehavior. PTAL..

grafnu (approved):

PR #187, opened 2020-10-06
Do not raise exception in case of out-of-sequence Faucet event

didovesei:
Every time Forch connects to the socket it will restore the states from varzs. So I think we can remove the exception here.

### DIFF
--- a/testing/python_lib/test_event.py
+++ b/testing/python_lib/test_event.py
@@ -22,9 +22,17 @@ class FaucetEventOrderTestCase(ForchestratorTestBase):
 
     def _handle_connection(self, event_socket):
         event_socket.listen(1)
-        while True:
-            print('Waiting for incoming connection')
-            self._connection, _ = event_socket.accept()
+        print('Waiting for incoming connection')
+        self._connection, _ = event_socket.accept()
+
+        events = [
+            {'version': 1, 'time': 123.0, 'event_id': 101},
+            {'version': 1, 'time': 124.0, 'event_id': 200},
+        ]
+        for event in events:
+            print(f'Sending event: {event}')
+            event_bytes = bytes('\n'.join((json.dumps(event, default=str), '')).encode('UTF-8'))
+            self._event_server_connection.sendall(event_bytes)
 
     def _setup_event_server(self):
         assert self._temp_socket_file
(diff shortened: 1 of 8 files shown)


grafnu (commented):
Have the implications of this been sorted through with Ops & Eng?  Basically, with this PR the system will get into an inconsistent state when this happens... which really means that the process should be restarted or refreshed -- e.g. exit and full restart, or restore state from metrics.  Or, need explicit agreement from the team (eng/ops) that "it's ok that Forch will be inconsistent until somebody deals with the metric being incremented"

Basically, has the interstitial time from when this error occurs to something being done about it been thought about and clarified?

didovesei:
> which really means that the process should be restarted or refreshed -- e.g. exit and full restart, or restore state from metrics

I agree that Eng/Ops should have been engaged. Just a clarification before involving them. I think we are already restoring the states from metrics. Do we still need to restart Forch or Faucet when there is an out-of-sequence event, in order for Forch to have states consitent with Faucet?

grafnu:
Hmmm.... thinking about this more, maybe it wasn't necessary to engage them
previously if we have a clean solution in mind -- I was operating under the
assumption that things were *not* automatically restored when this happens.
But, I think this means two things:

   - We need some kind of test to make sure that when this condition
   happens that state is restored. Not 100% sure how to do this, but, it
   likely will require an artificial (not Faucet) event source that can be
   primed to feed events that trigger this, and then some way to check that
   the "state restored" bit of code happened. Have a think about how to do
   this and we can see what makes sense... it's not obvious to me what the
   right path here would be (how to test it without excessive work).
   - Really then we just need to let them know saying "this condition
   should be handled internally to Faucet/Forch now, but we created this new
   metric to monitor if it's happening" -- not sure if there's anything they
   would do on their end with that.

If the state isn't restored internally to forch, then we need to either
make it so, or go the other route.  If it is restored, then we don't need
to restart (since it's functionally equivalent to restoring).

grafnu:
Might be that it's sufficient in a test to inject a single bad event (so
sequence number will be bad), and then make sure that the "restore state"
function is called... Seems like that should be a simple unit test -- if
not, we should likely figure out what to do to make it a simple unit test!

didovesei:
> Might be that it's sufficient in a test to inject a single bad event (so sequence number will be bad), and then make sure that the "restore state" function is called... Seems like that should be a simple unit test -- if not, we should likely figure out what to do to make it a simple unit test!

Added unit test in "test_event.py" to test if "restore_states()" is called in case of out-of-sequence event.

testing/python_lib/test_event.py
```diff
@@ -0,0 +1,78 @@
...
+            self._forchestrator._faucet_events_connect()
+        except Exception as error:
+            print(f'Ignoring expected exception during restoring states: {error}')
+
+        self._forchestrator._faucet_events.set_event_horizon(100)
+        restore_states_called = False
+
+        try:
+            self._forchestrator.main_loop()
+        except InvalidConfigError as error:
```
grafnu, line 69: can you please add a comment explaining the logic here. I think it's that calling restore_states generates some other kind of error since the system is not properly configured?  Just a simple one line description of WHY this is would be great.  Otherwise, InvalidConfigError != FaucetEventOrderError so it seems off.

grafnu (commented):
One minor comment to add a comment... otherwise, great way to test this!

testing/python_lib/test_event.py
didovesei: Yes,  restore_states() will generate InvalidConfigError since Faucet config file is empty. Added the explanation in the comment..

grafnu (approved):

PR #188, opened 2020-10-07
Fot DHCP reflection changes

anurag6:


testing/python_lib/integration_base.py
```diff
@@ -118,8 +126,12 @@ def _write_faucet_config(self, config):
         return self._write_yaml_to_file(filename, config)
 
     def _get_faucet_config_path(self):
-        return os.path.dirname(os.path.abspath(__file__)) + \
-            '/../../inst/forch-faucet-1/faucet/faucet.yaml'
+        if self.stack_options.get('fot'):
+            return os.path.dirname(os.path.abspath(__file__)) + \
+                '/../../inst/forch-controller-1/faucet/faucet.yaml'
```
grafnu, line 131: replace common string with format placeholder constant (../inst/%s/) and then use % substitution

testing/python_lib/test_fot.py
```diff
@@ -215,5 +216,40 @@ def test_ports_states(self):
...
 
+class FotContainerTest(IntegrationTestBase):
+    """Test suite for dynamic config changes"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.stack_options['fot'] = True
+        self.stack_options['local'] = True
+
+    def tearDown(self):
```
grafnu: remove function entirely

etc/FAUCET_VERSION
```diff
@@ -1 +1 @@
-1.9.48
+master
```
grafnu, line 1: I think the requisite change is in 1.9.50 that's currently available, so if that works we should just update here and make sure it all still works. I'm in the middle of testing 1.9.50 integration and it all seems good, but might as well get that tested with this too.

etc/start_faux
```diff
@@ -2,6 +2,8 @@
 
 echo Running faux $*
 
+umount /etc/resolv.conf
```
grafnu, line 5: add comment as to WHY this is here.

testing/python_lib/test_fot.py
```diff
@@ -215,5 +216,40 @@ def test_ports_states(self):
...
+        pass
+
+    def test_dhcp_reflection(self):
+        """Test to check DHCP reflection when on test VLAN"""
+        def dhclient_method(container=None):
```
grafnu, line 229: Either use positional args or keyword, but try not to mix them. E.g., here you're using a keyword arg (container=), but when you call it you're just using positional i.e. dhclient_metho('forch-faux-1')  -- just need to fix one of the two places to match the other.

grafnu (commented):
Nice... getting close!

testing/python_lib/integration_base.py
anurag6: Addressed.

testing/python_lib/test_fot.py
anurag6: Done.

etc/FAUCET_VERSION
anurag6: changed to 1.9.50

etc/start_faux
anurag6: Done

testing/python_lib/test_fot.py
anurag6: Done

testing/python_lib/integration_base.py
```diff
@@ -128,10 +130,9 @@ def _write_faucet_config(self, config):
     def _get_faucet_config_path(self):
         if self.stack_options.get('fot'):
             return os.path.dirname(os.path.abspath(__file__)) + \
-                '/../../inst/forch-controller-1/faucet/faucet.yaml'
-        else:
-            return os.path.dirname(os.path.abspath(__file__)) + \
-                '/../../inst/forch-faucet-1/faucet/faucet.yaml'
+                ('/../../inst/%s/faucet/faucet.yaml' % ('forch-controller-1'))
```
grafnu, line 133: extract shared string to constant

etc/start_faux
```diff
@@ -2,7 +2,7 @@
 
 echo Running faux $*
 
-umount /etc/resolv.conf
+umount /etc/resolv.conf # Necessary to run dhclient in docker containers
```
grafnu, line 5: comment goes on previous line not trailing (just style convention)

anurag6: Resolved.

testing/python_lib/integration_base.py
anurag6: changed.

etc/start_faux
```diff
@@ -2,8 +2,8 @@
 
 echo Running faux $*
 
-umount /etc/resolv.conf # Necessary to run dhclient in docker containers
-
+# Necessary to run dhclient in docker containers
+umount /etc/resolv.conf
```
grafnu, line 6: now you lost the blank space after - it's a separate logical construct from the things that follow it

testing/python_lib/integration_base.py
```diff
@@ -128,11 +128,12 @@ def _write_faucet_config(self, config):
         return self._write_yaml_to_file(filename, config)
 
     def _get_faucet_config_path(self):
+        config_file = '/../../inst/%s/faucet/faucet.yaml'
```
grafnu, line 131: config_file_format

etc/start_faux
anurag6: changed

testing/python_lib/integration_base.py
anurag6: changed.

grafnu (approved):

PR #190, opened 2020-10-07
Update NOAH APIs to show a list of IPs for a host

didovesei:


proto/switch_state.proto
```diff
@@ -119,8 +119,8 @@ message SwitchState {
     // Port number on this switch
     int32 port = 1;
 
-    // Learned IP address for this MAC
-    string ip_address = 2;
+    // Learned IP addresses for this MAC
+    repeated string ip_addresses = 2;
```
grafnu, line 123: I think we should make this mac_ips for consistency with the others (src_ips, dst_ips, host_ips)

grafnu (commented):
And just to clarify, under what conditions is the list of ips (for various bits) cleared?  What keeps the list of ip addresses from becoming arbitrarily long?

proto/switch_state.proto
didovesei: Updated.

Regarding the clearance, it will be cleared when the MAC expires. In this case, the entire ip list will be removed along with other information of the MAC.

grafnu (approved):

PR #191, opened 2020-10-13
Change Content-Type in response headers for varz page

anurag6:


forch/http_server.py
```diff
@@ -45,6 +45,8 @@ def do_GET(self):
                 opts[pair[0]] = pair[1]
             message = str(self._context.get_data(self.headers.get('Host'), parsed.path[1:], opts))
             self.send_response(http.HTTPStatus.OK)
+            if self._context.content_type:
+                self.send_header('Content-type', self._context.content_type)
```
grafnu, line 49: Does this apply to everything, or only varz requests?

anurag6: Yup. Not for the normal forch page

grafnu: So how does it not get applied to everything?  From a quick read, it seems that it will get applied to everything that goes through the proxy.  How is it limited to varz pages?

forch/forch_proxy.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, proxy_config):
     def start(self):
         """Start proxy server"""
         self._register_pages()
-        self._proxy_server = HttpServer(self._proxy_port)
+        self._proxy_server = HttpServer(self._proxy_port, content_type='text/plain')
```
grafnu, line 27: add a comment here saying that forch_proxy is only currently used for varz so this server can be hard-wired to text/plain

anurag6: Renamed proxy instance variable instead to clarify that the instance is only used for varz

grafnu (approved):

PR #192, opened 2020-10-14
Validate mirror and trunk ports.

anurag6:


testing/python_lib/unit_base.py
```diff
@@ -313,3 +314,28 @@ def _verify_ports_states(self, expected_states):
 
     def _verify_received_device_behaviors(self, expected_device_behaviors):
         self.assertEqual(self._received_device_behaviors, expected_device_behaviors)
+
+
+class ForchestratorTestBase(UnitTestBase):
```
grafnu, line 319: Somebody's going to have a conflict! -- https://github.com/faucetsdn/forch/pull/187

I think it makes sense to have one class for this, just will need to resolve/combine functionality.  Who ever gets there first wins!

grafnu (approved):

PR #193, opened 2020-10-14
Reset timeout values for test_scale and test_reconfig

anurag6:


.github/workflows/main.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
 
   test_reconfig:
     runs-on: ubuntu-latest
-    timeout-minutes: 40
+    timeout-minutes: 50
```
grafnu, line 67: are you sure this is needed?  I'm seeing a successful run of this take ~20min, why bump to 50?

grafnu (commented):
Thanks for looking at this!  I know test_scale needs it, but I didn't think test_reconfig did... are you sure that's necessary?

.github/workflows/main.yaml
pbatta: test_reconfig also timed out a few times (https://github.com/faucetsdn/forch/actions/runs/305295427, https://github.com/faucetsdn/forch/actions/runs/305122643).

grafnu:
Yes, but the question is if the increased-with-longer-timeout is a valid
success condition -- or is it wedged indefinitely? Or indicative of a
failure condition?  And/or, why is there such high variance in the success
time?

anurag6:
> Thanks for looking at this! I know test_scale needs it, but I didn't think test_reconfig did... are you sure that's necessary?

Oh I jst wanted to do a sample PR to see if @pbatta hypothesis works.

anurag6:
I am running the tests a few times to see if any of them fails. test_reconfig doesn't seem to be wedged indefinitely, though. It was usually at different points of the run where it was canceled. If it were edged indefinitely, the additional 10 min bump shouldn't help it. We could run the tests a few times and see if it fails again.

PR #198, opened 2020-10-21
WIP: Change egress link to 1 in sim setups

anurag6:


grafnu (approved):
(Pending tests passing, of course).

Just curious -- why this change? I have no problem with it just wondering what the triggering motivation was...   The only "concern" would to be to maintain parity with the actual production layouts, but that's a secondary consideration.

anurag6:
> (Pending tests passing, of course).
> 
> Just curious -- why this change? I have no problem with it just wondering what the triggering motivation was... The only "concern" would to be to maintain parity with the actual production layouts, but that's a secondary consideration.

I am sorry, this was just the first commit for the scaling PR. I wanted to see if tests go through. The idea is to limit all links that dont go down to t2 switches in the range 1-9. The other links will be used purely for t2 scaling.

PR #203, opened 2020-10-26
Add scheduling for config hash verficiation

didovesei:
Making Forch retry config hash match checking and not raise an exception until a limit is reached.


forch/forchestrator.py
```diff
@@ -45,6 +45,7 @@
 _FAUCET_PROM_PORT_DEFAULT = 9302
 _GAUGE_PROM_HOST = '127.0.0.1'
 _GAUGE_PROM_PORT_DEFAULT = 9303
+_CONFIG_HASH_MAX_RETRY = 5
```
grafnu, line 48: make configurable (externally) with default to 5

forch/forchestrator.py
```diff
@@ -113,6 +114,9 @@ def __init__(self, config):
         self._metrics = None
         self._varz_proxy = None
 
+        self._config_hash_retry = 0
+        self._config_hash_max_retry = os.getenv('_CONFIG_HASH_MAX_RETRY', _CONFIG_HASH_MAX_RETRY)
```
grafnu, line 118: why env variable rather than config param?

forch/forchestrator.py
```diff
@@ -436,7 +440,15 @@ def _restore_states(self):
...
         config_info, faucet_dps, _ = self._get_faucet_config()
         self._update_config_warning_varz()
-        assert config_hash == config_info['hashes'], 'config hash info does not match'
+
+        if config_hash == config_info['hashes']:
+            self._config_hash_retry = 0
+        else:
+            assert self._config_hash_retry < self._config_hash_max_retry, (
+                f'config hash info does not match after {self._config_hash_max_retry} retries')
+            self._config_hash_retry += 1
```
grafnu, line 449: how often does this happen?  What, in practical terms, does a value of "5" mean?  Is it periodic or 5 of some particular kind of event?

forch/forchestrator.py
didovesei: Added _DEFAULT

forch/forchestrator.py
didovesei: I thought we should hide this implementation from the user, but I think we do need to expose it in Forch config in case the default value is not large enough.

forch/forchestrator.py
didovesei: I just ran Eng test again and analyzed the sequence of the events. I think this value is mostly about how offen there is an L2_EXPIRE event is received with a CONFIG_CHANGE event. And that is related to the number of devices and the timing of the expiry of the devices. For example, if every few seconds there is a device expires, it is more likely that a hash mismatch happens, because Faucet reloads the config every few seconds.

In Eng's test, I can see the highest retry is 3 with about 10 devices.

forch/forchestrator.py
grafnu: But hide-or-not-hide is not env-vs-config... they are the same w.r.t. if they exist or not.

forch/forchestrator.py
grafnu: So -- do you think a simple count of the number of hash failures is the right function?  Seems fairly fragile and dependent on the number of devices, etc... Seems like a better thing would be "X seconds after the last hash clash"?

proto/forch_configuration.proto
```diff
@@ -237,3 +240,11 @@ message DataplaneMonitoring {
   // threshold of number of packets per second for each vlan; indexed by vlan
   map<int32, int32> vlan_pkt_per_sec_thresholds = 2;
 }
+
+/*
+ * Parameters for Faucet config porcessing
+ */
+message FaucetConfigProcessing {
+  // maximum retry times for Faucet config hash checking
+  int32 config_hash_max_retry = 1;
```
grafnu, line 249: can/should this be put in the EventClientConfig message?  Seems like it's conceptually related...

forch/forchestrator.py
didovesei: Got it. Added `config_hash_cooling_sec` within `EventClientConfig` in forch config.

forch/forchestrator.py
didovesei: Changed to a time based...

proto/forch_configuration.proto
didovesei: Yes, moved to EventClientConfig..

forch/forchestrator.py
```diff
@@ -45,7 +45,7 @@
 _FAUCET_PROM_PORT_DEFAULT = 9302
 _GAUGE_PROM_HOST = '127.0.0.1'
 _GAUGE_PROM_PORT_DEFAULT = 9303
-_CONFIG_HASH_MAX_RETRY_DEFAULT = '5'
+_CONFIG_HASH_COOLING_SEC_DEFAULT = '30'
```
grafnu, line 48: COOLING?  Not sure that's right -- timeout would be more appropriate.

forch/forchestrator.py
```diff
@@ -114,10 +114,10 @@ def __init__(self, config):
         self._metrics = None
         self._varz_proxy = None
 
-        self._config_hash_retry = 0
-        self._config_hash_max_retry = (
-            self._config.faucet_config_processing.config_hash_max_retry or
-            int(os.getenv('_CONFIG_HASH_MAX_RETRY', _CONFIG_HASH_MAX_RETRY_DEFAULT))
+        self._config_hash_clash_start_time = None
+        self._config_hash_cooling_sec = (
```
grafnu, line 118: consistency in names -- the prefix should be common up to the point where they are talking about different things. Specificaly,  _config_hash_clash and _config_hash_cooling  should be the same since they are referring to the same concept (time of the config hash mismatch)

grafnu (commented):
Still not sure about the intended algorithm. What happens if there's a single hash mismatch event and then nothing else?

forch/forchestrator.py
didovesei: Changed to TIMEOUT.

forch/forchestrator.py
didovesei: Updated. Using `self._config_hash_clash_start_time` and `self._config_hash_clash_timeout_sec` now..

grafnu (commented):
Still not sure about the intended algorithm. What happens if there's a single hash mismatch event and then nothing else?

didovesei:
> Still not sure about the intended algorithm. What happens if there's a single hash mismatch event and then nothing else?

So the reason there was a config hash mismatch is that the behavioral Faucet config file was updated by Forch while Forch received a CONFIG_CHANGE event. So once there is a config hash mismatch, Faucet will reload the config file and send a new CONFIG_CHANGE event. If the hashes match, then Forch will clear the `config_hash_clash_start_time`. Otherwise, it means Forch updated the config again, and `config_hash_clash_start_time` won't be cleared until a CONFIG_CHANGE event is received that makes the hashes match.

I think the key assumption is that Faucet will always reload the config file if it is different and emit a CONFIG_CHANGE event. I mean, there will be always another CONFIG_CHANGE event after a single hash mismatch till the hashes match...

grafnu:
"there will be always another CONFIG_CHANGE event after a single hash
mismatch till the hashes match" -- how do you know that? Faucet itself
doesn't know about the mismatch, it just knows that the config changed, so
it won't categorically send another config change after a hash failure. The
sequence of events you cite is the reason for *this* mismatch, but isn't
the reason for the error check in the first place.  It's a distributed
system, and so there should be some mechanism to ensure eventual
consistency in the face of weird behavior.   What happens if there's some
calculation error for some unknown reason on the Faucet side, and Forch
doesn't update the config again b/c there was no additional change?  You'll
get a CONFIG_CHANGE event, the hashes will mismatch, and then... nothing.

forch/forchestrator.py
```diff
@@ -113,6 +114,14 @@ def __init__(self, config):
         self._metrics = None
         self._varz_proxy = None
 
+        self._config_hash_clash_timer = None
+        self._config_hash_clashed = False
+        self._config_hash_clash_timeout_sec = (
+            self._config.event_client.config_hash_clash_timeout_sec or
+            int(os.getenv(
```
grafnu, line 121: why env variable?

forch/forchestrator.py
```diff
@@ -436,9 +445,43 @@ def _restore_states(self):
     def _restore_faucet_config(self, timestamp, config_hash):
         config_info, faucet_dps, _ = self._get_faucet_config()
         self._update_config_warning_varz()
-        assert config_hash == config_info['hashes'], 'config hash info does not match'
+
+        if config_hash == config_info['hashes']:
+            self._attempt_cancel_config_hash_clash_timer()
```
grafnu, line 450: I think in these cases I'd leave out the "attempt" -- it's not semantically meaningful

forch/forchestrator.py
```diff
@@ -436,9 +445,43 @@ def _restore_states(self):
...
+        else:
+            LOGGER.warning('Config hash does not match')
+            self._attempt_start_config_hash_clash_timer()
+
         self._faucet_collector.process_dataplane_config_change(timestamp, faucet_dps)
 
+    def _attempt_start_config_hash_clash_timer(self):
+        if self._config_hash_clash_timer and self._config_hash_clash_timer.is_alive():
+            return
+        self._config_hash_clash_timer = threading.Timer(
```
grafnu, line 460: you should reuse one of the _existing_ heartbeat schedulers for this -- creating lots of little threads ends up being very messy in the long run.

forch/forchestrator.py
```diff
@@ -436,9 +445,43 @@ def _restore_states(self):
...
+
+        if config_hash == config_info['hashes']:
+            self._attempt_cancel_config_hash_clash_timer()
+        else:
+            LOGGER.warning('Config hash does not match')
+            self._attempt_start_config_hash_clash_timer()
+
         self._faucet_collector.process_dataplane_config_change(timestamp, faucet_dps)
 
+    def _attempt_start_config_hash_clash_timer(self):
```
grafnu, line 457: this shouldn't be "start a timer" -- but rather just a mark of when the last config hash clash was

forch/forchestrator.py
```diff
@@ -436,9 +445,43 @@ def _restore_states(self):
...
+    def _check_config_hash_clashed(self):
+        _, varz_config_hashes = self._get_varz_config()
+        config_info, _, _ = self._get_faucet_config()
+
+        if varz_config_hashes != config_info['hashes']:
+            LOGGER.error(
+                'Config hash does not match after %s seconds', self._config_hash_clash_timeout_sec)
+            self._config_hash_clashed = True
+
+    def _get_config_hash_clashed(self):
```
grafnu, line 482: private getter is redundant -- just use the variable directly

forch/faucet_event_client.py
```diff
@@ -212,9 +212,11 @@ def _dispatch_faucet_event(self, target, target_event):
     def _should_log_event(self, event):
         return event and os.getenv('FAUCET_EVENT_DEBUG')
 
-    def next_event(self, blocking=False):
+    def next_event(self, get_config_hash_clashed, blocking=False):
```
grafnu, line 215: this is not right -- the concept of a config hash shouldn't bleed down into the event client itself

grafnu (commented):
> Added timer which starts when Forch detects the every first hash mismatch.
> 
> Also comparing the hash again when timer times out.. The reason is that while running Eng's test, it happened that Forch wrote behavioral config in between Faucet reloading. For example what happened was:
> T1.
> 
> 1. Faucet reloaded config version 1
> 2. Forch received an L2_EXPIRE event
> 3. Forch wrote out behavior config version 2 due to the expiry event
> 4. Forch received a CONFIG_CHANGE event with hash version 1
> 
> T2.
> 
> 1. Forch recived learning event, wrote out behavior config version 1
> 
> T3.
> 
> 1. Faucet reloaded config version 1, did nothing because config is unchanged from its perspective

I think the right answer here is that the timeout should be reset when forch writes a new config.    It's a valid thing for forch to continually write out new configs, in which case the timer likely shouldn't expire. It's the other extreme case (form only one event).

forch/forchestrator.py
didovesei: Removed..

forch/forchestrator.py
didovesei: Not using a seprate timer now.

forch/forchestrator.py
didovesei: Using _faucet_state_scheduler now..

forch/forchestrator.py
didovesei: removed

forch/faucet_event_client.py
didovesei: removed

didovesei:
> > Added timer which starts when Forch detects the every first hash mismatch.
> > Also comparing the hash again when timer times out.. The reason is that while running Eng's test, it happened that Forch wrote behavioral config in between Faucet reloading. For example what happened was:
> > T1.
> > 
> > 1. Faucet reloaded config version 1
> > 2. Forch received an L2_EXPIRE event
> > 3. Forch wrote out behavior config version 2 due to the expiry event
> > 4. Forch received a CONFIG_CHANGE event with hash version 1
> > 
> > T2.
> > 
> > 1. Forch recived learning event, wrote out behavior config version 1
> > 
> > T3.
> > 
> > 1. Faucet reloaded config version 1, did nothing because config is unchanged from its perspective
> 
> I think the right answer here is that the timeout should be reset when forch writes a new config. It's a valid thing for forch to continually write out new configs, in which case the timer likely shouldn't expire. It's the other extreme case (form only one event).

I have run the test again and it seems that simply resetting the timer (setting clash_start_time to current time) could not solve the issue... The problem is that the timer is only cleared when there is a CONFIG_CHANGE event. In the test's case, there was not a new CONFIG_CHANGE event, as the file got changed by Forch to the previous version. I guess we need to compare the hashes when the timer times out, and clear the timer (setting clash_start_time to None) if the hashes match..

grafnu:
I'll look at the PR now, but the timer should be cleared *when forch writes
a new config.  * So -- as long as forch *knows* it's updating the config
then there's no need for the error, it's only when forch *thinks* it's
settled and there's still an error do we need to flag the condition.

grafnu:
Ah, I see what you're saying now about how faucet doesn't send a new
update.   Would something like this work, then, where when the time after
an intentional config change expires, the system compares the "last
received hash" with the "expected hash" -- and it's an error at that
point?  Maybe that's getting back to what you were saying about needing to
recalculate at that point (really you just need to recalculate when it is
written out, but that can be done at the time of write or the time of
check).  So basically the logic becomes:

   1. Whenever forch writes a config, start/reset the timer.
   2. After XXX seconds, check to see if the last received config hash
   (from faucet) matches the last written config hash (from step #1).  If
   there's a mismatch, then error.

Does that work for the cases you've been tracking?

didovesei:
> something like this work, then, where when the time after an intentional config chang

Yes, I think that should work. So we do not need to start the timer when we detect a mismatch anymore,  right? The new logic seems to be: "After Forch writes a config, within XXX seconds, we expect Faucet to reload it and apply the changes, and the new hashes should match. Or, if Faucet sees the config as the same with the previous one, the hash of the file should match the last hash sent out by Faucet".

grafnu:
Correct... The triggering event is the config write, not the bad hash. Also
covers the case where forch writes a config, but there is never a response.

didovesei:
> Correct... The triggering event is the config write, not the bad hash. Also
covers the case where forch writes a config, but there is never a response.

updated the PR. PTAL..

grafnu (approved):
Please update title (no longer max retry times)

Thanks for working on this -- I know it was a lot of back and forth, but in the end I think we got to the right solution that's more robust than before (e.g., the case of "no config event at all" was not checked previously!)

PR #207, opened 2020-11-02
Cleanup controller setup for latest DAQ/FOT integration

grafnu:


grafnu:
Note this is blocked by integration of Faucet 1.9.51 that is both required to make this work (allowing IN_PORT specification) but also breaks something else (forch stack test).

grafnu:
Closing to shift to a pull against fot not master.

PR #208, opened 2020-11-04
Determine CPN node state based on consecutive successful ping counts

didovesei:


didovesei:
Making Forch ping the node every `x` seconds. If Forch receives `y` consecutive successful ping replies from a node, then it considers it as healty. If Forch receives `z` consecutive ping failures, then it considers it as down. In all the other cases, it considers the node as damaged.

x, y, z are configurable parameters. The default values are 2, 3, 10, respectively.

grafnu (approved):

PR #210, opened 2020-11-06
Tweaks and configs to enable Forch

didovesei:


grafnu (approved):
Can you try deleting the faucet_br directory for topo/bond and see if all the other tests pass? I don't think it's used by anything else so we should remove it if not used.

PR #211, opened 2020-11-06
Parallelize connectivity tests and faux device creation to boost setup speed.

anurag6:


testing/python_lib/test_failscale.py
```diff
@@ -34,14 +34,21 @@ def __init__(self, *args, **kwargs):
             'skip-conn-check': False
         })
 
+    def tearDown(self):
```
grafnu: what's this for?  (i.e., why not let the superclass handle it?)

testing/python_lib/test_failscale.py
```diff
@@ -34,14 +34,21 @@ def __init__(self, *args, **kwargs):
...
+        faux_args = []
         for dev_index in range(1, self.devices):
             for sw_index in range(1, self.switches + 1):
                 device_num = dev_index * self.switches + sw_index
                 switch = 't2sw%s' % sw_index
-                self.add_faux(switch, 100 + dev_index + 1, device_num)
+                faux_args.append((switch, 100 + dev_index + 1, device_num))
+
+        self.parallelize(len(faux_args), self.add_faux, faux_args)
+#                self.add_faux(switch, 100 + dev_index + 1, device_num)
```
grafnu: ?

testing/python_lib/integration_base.py
```diff
@@ -114,7 +116,8 @@ def _get_docker_ip(self, container, interface='faux-eth0'):
         _, out, _ = self._run_cmd('ip addr show %s' % (interface),
                                   docker_container=container, capture=True)
         out_list = out.split()
-        return out_list[out_list.index('inet') + 1].split('/')[0]
+        return out_list[out_list.index('inet') + 1].split('/')[0] \
```
grafnu, line 119: don't use continuation, create local variable if necessary

testing/python_lib/integration_base.py
```diff
@@ -141,6 +144,40 @@ def _get_faucet_config_path(self):
         return os.path.dirname(os.path.abspath(__file__)) + \
             (config_file_format % ('forch-faucet-1'))
 
+    def parallelize(self, job_count, target, target_args=None):
+        """Parallelizes multiple runs of a target method with multiprocessing"""
+        jobs = []
+        if not target_args:
+            target_args = []
+        for job in range(job_count):
+            try:
```
grafnu, line 153: rather than catching the exception, do an explicit if check with the length

testing/python_lib/test_failscale.py
```diff
@@ -26,7 +27,7 @@ def __init__(self, *args, **kwargs):
             yaml.dump(config, config_file)
 
         self.stack_options.update({
-            'devices': self.devices,
+            'devices': 1,
```
grafnu, line 30: This doesn't seem right... how is self.devices manifest?  And if the answer is "elsewhere" -- why is this one and not zero?  I'd expect all the devices to be created in one place.

testing/python_lib/test_failscale.py
```diff
@@ -35,23 +36,50 @@ def __init__(self, *args, **kwargs):
 
     def test_stack_connectivity(self):
         """Test to build stack and check for connectivity"""
+
+        faux_args = []
+        for dev_index in range(1, self.devices):
+            for sw_index in range(1, self.switches + 1):
+                device_num = dev_index * self.switches + sw_index
+                switch = 't2sw%s' % sw_index
+                faux_args.append((switch, 100 + dev_index + 1, device_num))
```
grafnu, line 45: extract "100 + ..." to named variable so it's obvious what it is (device_port?)

testing/python_lib/test_failscale.py
```diff
@@ -35,23 +36,50 @@ def __init__(self, *args, **kwargs):
...
-                                 'Couldn\'t reach 192.168.1.1 from %s' % device)
+            ping_count = multiprocessing.Array("i", [0, 1])
+            ping_counts[device] = ping_count
+            target_args.append((device, ping_count, ))
+        self.parallelize(len(target_args), ping_device, target_args)
+
+        for device in ping_counts:
+            print('Device %s ping_count: %s' % (device, ping_counts[device][:]))
+        for device in ping_counts:
+            self.assertLessEqual(1, ping_counts[device][0], 'Device %s \
```
grafnu, line 66: avoid continuation, break line elsewhere (e.g. after comma, etc...)

testing/python_lib/test_failscale.py
```diff
@@ -35,23 +36,50 @@ def __init__(self, *args, **kwargs):
...
-        time.sleep(10)
+        time.sleep(3)
         self._fail_egress_link()
         try:
             ping_count = self._ping_host_reap(process, output=True)
             # Check that at least some flow was disrupted, but not too much.
-            self.assertTrue(2 < ping_count < 39, 'disrupted ping count %s' % ping_count)
+            # Note: Changing it to greater than equal 40 since on higher devices,
+            # there are cases where there is no disruption observed.
+            self.assertTrue(2 < ping_count <= 40, 'disrupted ping count %s' % ping_count)
```
grafnu, line 82: not sure this is right -- there are cases where we want to make sure there is a non-zero amount of disruption...  so, might need to qualify this (e.g., upped_bound == 39 unless deive_num > 10 or something like that).

testing/python_lib/test_failscale.py
```diff
@@ -35,23 +36,50 @@ def __init__(self, *args, **kwargs):
...
+            ping_count[1] = self._ping_host(device, '192.168.1.1', count=10, output=True)
+
+        ping_counts = {}
+        target_args = []
         for device in device_list:
-            self.assertLessEqual(1, self._ping_host(device, '192.168.1.0', count=3, output=True),
-                                 'Couldn\'t reach 192.168.1.0 from %s' % device)
-            self.assertLessEqual(1, self._ping_host(device, '192.168.1.1', count=3, output=True),
-                                 'Couldn\'t reach 192.168.1.1 from %s' % device)
+            ping_count = multiprocessing.Array("i", [0, 1])
```
grafnu, line 58: this should be abstracted away into the superclass somehow -- there should be no need to reference or understand "multiprocessing" in the context of test_failscale.py

testing/python_lib/test_failscale.py
```diff
@@ -1,6 +1,7 @@
 """Integration test base class for Forch"""
 import unittest
 import time
+import multiprocessing
```
grafnu, line 4: as per comment below, there should be no mention of "multiprocessing" in this file -- everything specific to that should be abstracted away somehow

testing/python_lib/test_failscale.py
anurag6: Outdated code

testing/python_lib/test_failscale.py
anurag6: Outdated code

testing/python_lib/integration_base.py
anurag6: Changed.

testing/python_lib/integration_base.py
anurag6: Changed

testing/python_lib/test_failscale.py
anurag6: Changed to remove devices option in test_scale itself.

testing/python_lib/test_failscale.py
anurag6: Changed

testing/python_lib/test_failscale.py
anurag6: Changed

testing/python_lib/test_failscale.py
anurag6: Changed upperbound to 39

testing/python_lib/test_failscale.py
anurag6: Changed

testing/python_lib/test_failscale.py
anurag6: Changed

testing/python_lib/test_failscale.py
```diff
@@ -59,18 +56,20 @@ def ping_device(device, ping_count):
         ping_counts = {}
         target_args = []
         for device in device_list:
-            ping_count = multiprocessing.Array("i", [0, 1])
+            ping_count = self.get_multiprocessing_array('i', 2)
```
grafnu, line 59: not quite what I meant, all you did was rename things, but the _concept_ of multiprocessing is still there -- it's bleeding the abstraction.  Case in point -- what is this multiprocessing array anyway?  Looking at this makes no sense to me (without a detailed understanding of the multiprocessing something something) -- what is i? what is 2?

The real underlying problem here is that the code is not obvious what it's doing. I suppose the other way out of this would be to comment it to explain what it's doing... but in a better world you wouldn't have to do that (documentation is an admission of failure).

I suppose I can't even suggest what the proper fix is because I don't understand what the code is trying to do...  So Maybe start with a comment or two explaining what this array is, and then we can either stick with that or I could suggest a better refactoring.

anurag6: Does it make sense now?

testing/python_lib/test_failscale.py
```diff
@@ -35,23 +34,55 @@ def __init__(self, *args, **kwargs):
...
+
+        ping_counts = {}
+        target_args = []
         for device in device_list:
-            self.assertLessEqual(1, self._ping_host(device, '192.168.1.0', count=3, output=True),
-                                 'Couldn\'t reach 192.168.1.0 from %s' % device)
-            self.assertLessEqual(1, self._ping_host(device, '192.168.1.1', count=3, output=True),
-                                 'Couldn\'t reach 192.168.1.1 from %s' % device)
+            # Get shared memory int array which ensures same copy of the array across
+            # multiple processes
```
grafnu, line 60: if it's shared across multiple processes why is it allocated once for each one?  Something about this doesn't smell right, so either it's confused, or I'm confused. If we can't figure it out in the comments then maybe we'll need a quick GVC to clarify...

testing/python_lib/integration_base.py
```diff
@@ -141,6 +145,44 @@ def _get_faucet_config_path(self):
         return os.path.dirname(os.path.abspath(__file__)) + \
             (config_file_format % ('forch-faucet-1'))
 
+    def parallelize(self, job_count, target, target_args=None):
+        """Parallelizes multiple runs of a target method with multiprocessing"""
+        jobs = []
+        if not target_args:
+            target_args = []
+        for job in range(job_count):
+            if job < len(target_args):
```
grafnu, line 154: is this complexity even necessary?  Seems like in all cases you just need the target_args and target, the "job_count" field is redundant?  Seems like this method could be simplified by removing the "job_count" argument and always having job_count=len(target_args)

anurag6: But what if the method has no arguments. Having to push an array with a set of empty tuples seems wrong.

grafnu:
"what if"

When is that gonna happen?

testing/python_lib/integration_base.py
anurag6: Changed.

testing/python_lib/test_failscale.py
```diff
@@ -35,23 +34,55 @@ def __init__(self, *args, **kwargs):
...
+            ping_count[1] = self._ping_host(device, '192.168.1.1', count=10, output=True)
+
+        ping_counts = {}
+        target_args = []
         for device in device_list:
-            self.assertLessEqual(1, self._ping_host(device, '192.168.1.0', count=3, output=True),
-                                 'Couldn\'t reach 192.168.1.0 from %s' % device)
-            self.assertLessEqual(1, self._ping_host(device, '192.168.1.1', count=3, output=True),
-                                 'Couldn\'t reach 192.168.1.1 from %s' % device)
+            # Get shared memory int array which ensures same copy of the array across
```
grafnu, line 59: nit: ensures is not the word you want.  Basically can just say "Use shared memory array between parent and child process"

grafnu (approved):
Just one minor comment suggestion.

PR #216, opened 2020-11-17
Enable Forch and DAQ reporting in FOT testing

didovesei:


grafnu:
test_fot failed this time around, so I restarted the jobs to see if it was transient

bin/setup_base
```diff
@@ -9,7 +9,7 @@ GO_VERSION=1.13.3
 sudo apt-get update
 sudo apt-get install apt-transport-https ca-certificates curl gnupg-agent software-properties-common
 curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
-sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/$(lsb_release -is | awk '{print tolower($0)}') $(lsb_release -cs) stable"
```
grafnu, line 12: extract release to variable  (mabye the -cs one as well)

testing/test_fot
```diff
@@ -1,22 +1,80 @@
...
+
+# Forch startup checks
+# Check Forch has successfuly started
+fgrep -o "Starting Forchestrator" $INST_BASE/forch.log | tee -a $TEST_RESULTS
+
+$YQ '.dps."nz-kiwi-t2sw1".interfaces."1".native_vlan' $BEHAVIORAL_CONFIG | tee -a $TEST_RESULTS
+$YQ '.dps."nz-kiwi-t2sw2".interfaces."1".native_vlan' $BEHAVIORAL_CONFIG | tee -a $TEST_RESULTS
+
+# Start DHCP client on faux-1 and faux-2
+docker exec forch-faux-1 dhclient -r &
```
grafnu, line 61: this should be done with the appropriate arg passed to cmd/faux

testing/test_fot
```diff
@@ -1,22 +1,80 @@
...
+docker exec forch-faux-2 dhclient -r
+docker exec forch-faux-1 dhclient &
+docker exec forch-faux-2 dhclient
+docker exec forch-faux-1 ip addr show faux-eth0 | grep inet
+docker exec forch-faux-2 ip addr show faux-eth0 | grep inet
+
+vlan=$($YQ '.dps."nz-kiwi-t2sw1".interfaces."1".native_vlan' $BEHAVIORAL_CONFIG)
+echo $(( (vlan>271) + (vlan>276) )) | tee -a $TEST_RESULTS
```
grafnu, line 69: spaces around >

testing/test_fot
```diff
@@ -1,22 +1,80 @@
...
+vlan=$($YQ '.dps."nz-kiwi-t2sw2".interfaces."1".native_vlan' $BEHAVIORAL_CONFIG)
+echo $(( (vlan>271) + (vlan>276) )) | tee -a $TEST_RESULTS
+
+# Waiting for DAQ test to complete
+function report_result {
+    $YQ '.dps."nz-kiwi-t2sw1".interfaces."1".native_vlan' $BEHAVIORAL_CONFIG | tee -a $TEST_RESULTS
+    $YQ '.dps."nz-kiwi-t2sw2".interfaces."1".native_vlan' $BEHAVIORAL_CONFIG | tee -a $TEST_RESULTS
+    cat $DAQBASE/inst/result.log | sort | tee -a $TEST_RESULTS
+}
+set +x
```
grafnu, line 79: remove +x?

forch/faucet_state_collector.py
```diff
@@ -1378,6 +1379,11 @@ def _get_port_attributes(self, switch, port):
                 ret_attr['type'] = 'egress'
                 return ret_attr
 
+            sequester_port_desc = self._config.orchestration.sequester_config.port_description
+            if sequester_port_desc and sequester_port_desc in port_info.get('description', ''):
```
grafnu, line 1383: this seems unstable (e.g. 'tap' would also match 'taper') -- likely should be looking for equality

didovesei: Done

testing/test_fot
didovesei: done

bin/setup_base
didovesei: Done

grafnu:
I mean whatever is used to create the faux container... not exactly sure
what that is, but it should be part of the base configuration if DHCP is
started or not.

testing/test_fot
didovesei: Actually I think this can be skipped because the device has never done DHCP before.. I have removed them.

forch/faucet_state_collector.py
```diff
@@ -1380,7 +1380,7 @@ def _get_port_attributes(self, switch, port):
                 return ret_attr
 
             sequester_port_desc = self._config.orchestration.sequester_config.port_description
-            if sequester_port_desc and sequester_port_desc in port_info.get('description', ''):
+            if sequester_port_desc and sequester_port_desc == port_info.get('description', ''):
```
grafnu, line 1383: don't need the .get() with default (since it would default to None anyway)

testing/test_fot
```diff
@@ -58,17 +58,15 @@ $YQ '.dps."nz-kiwi-t2sw1".interfaces."1".native_vlan' $BEHAVIORAL_CONFIG | tee -
 $YQ '.dps."nz-kiwi-t2sw2".interfaces."1".native_vlan' $BEHAVIORAL_CONFIG | tee -a $TEST_RESULTS
 
 # Start DHCP client on faux-1 and faux-2
-docker exec forch-faux-1 dhclient -r &
-docker exec forch-faux-2 dhclient -r
 docker exec forch-faux-1 dhclient &
```
grafnu, line 61: My comment was more about these -- dhclient should be run by a command line arg to the faux container startup, unless that's very difficult for some reason.

forch/faucet_state_collector.py
didovesei: Done

testing/test_fot
didovesei: Added faux_dhcp option for faux-1 and faux-2 in bin/setup_stack:145-146. This will start dhcp client during the faux container startup.

bin/stack_functions
```diff
@@ -103,10 +115,15 @@ function add_faux {
...
-    while ! docker exec forch-faux-$fnum ip addr show faux-eth0 | fgrep inet; do
+    for i in `seq 1 20`; do
         echo Waiting for IP address...
+        ip_address=$(docker exec forch-faux-$fnum ip addr show faux-eth0 | egrep -o '192\.168\.1\.[0-9]{1,3}' || true)
+        if [[ -n "$ip_address" ]]; then
+            break
+        fi
         sleep 2
     done
+    declare -g "ip_faux_$fnum"="$ip_address"
```
grafnu, line 126: ug -- this is a really messy way to do this. Do we know where this came from?  It's messy because it's "hidden" how the value gets set from the perspective of where it's used.

topo/fot/faucet/ryu.conf
```diff
@@ -0,0 +1,12 @@
+# Requires Ryu 4.0 or above, and allows quick detection of a down OpenFlow channel.
```
grafnu, line 1: were these replaced by topo/common?

bin/stack_functions
didovesei: So the main reason I added line 126 is that I could not find other ways to return the ip address to the caller of the add_faux function.. maybe I can just directly get the ip address after the function is called.. Will raise another PR.

grafnu: Also, the regex above is pretty ugly -- you're matching on a specific IP prefix which is fragile. I'd recommend something like "inet ([0-9.])/" with sed (as a capture group):

grafnu: `ip addr show wlp0s20f3 | sed -nr 's~inet ([0-9.]+)/.*~\1~p'`

PR #222, opened 2020-11-30
Set up testing with single t1 switch

didovesei:


grafnu (approved):
This is good for what it is (single t1), but we'll also need a separate setup (e.g. bin/setup_flat or something) that doesn't have the t1/t2 layering. Our initial test-case is really around having just a single switch, but then we might as well extend that to investigate having 2x+ access switches...

PR #223, opened 2020-12-01
Set up testing with flat toplogy

didovesei:
* Creating script to setup flat toplogy that consists of only access switches
** The number of switches is configurable
** The uplinks of the switches are bonded, and the switches themselves are connected in a ring topology

* Creating a test script that sets up the flat topology with one swith and test sconnectivity of devices and Forch APIs.


topo/flat/one/corp_conf.yaml
```diff
@@ -0,0 +1,19 @@
+dps:
```
grafnu, line 1: This should be parameterized, not different explicit topos for each case, there's simply too much redundancy -- that's just too unwieldy.  See testing/python_lib/build_config.py for the basics of that, but it should be extended to support these flat topologies as well.

grafnu (commented):
I think we need to parameterize the config files... it's too unwieldy to have all the different explicit configurations floating around.  I.e., modify/update testing/python_lib/build_config.py to do what's needed.

topo/flat/one/corp_conf.yaml
didovesei: Using build_config.py to generate Faucet configs for corp and faucet network now. PTAL..

didovesei:
> I think we need to parameterize the config files... it's too unwieldy to have all the different explicit configurations floating around. I.e., modify/update testing/python_lib/build_config.py to do what's needed.

Do we want to parameterize all the config files (ryu, gauge,forch, acls, etc.) or just the Faucet ones? Not sure if we want to parameterize ryu.conf and uniform.yaml since they are all identical in all the tests..

grafnu:
I'd have them stored in a "common" directory or "default" -- e.g.
topo/common/faucet/ryu.conf -- that is then included in all setups.  And
then, if a particular config has a need to change it we can deal with that
later...

didovesei:
> I'd have them stored in a "common" directory or "default" -- e.g. topo/common/faucet/ryu.conf -- that is then included in all setups. And then, if a particular config has a need to change it we can deal with that later...

Added topo/common for `ryu.conf`, `uniform.yaml` and `forch.yaml`. PTAL..

bin/setup_flat
```diff
@@ -35,15 +47,17 @@ for sw_num in $(seq 1 $switches); do
...
 
-    add_oeth $switch 50 $bonded_iface
+    add_oeth $switch 28 $bonded_iface
 done
 
 # set up ring links
-if (( $switches > 1)); then
-    for sw_num in $(seq 1 $switches); do
+if (( $switches > 2)); then
+    for sw_num in $(seq 1 $((switches-1))); do
```
grafnu, line 55: spaces around -

bin/setup_flat
```diff
@@ -72,7 +86,11 @@ container=forch-faucet-$inst
 container_inst=inst/$container/
 mkdir -p $container_inst/
 cp -r topo/$topo/faucet $container_inst/
+setup_flat_faucet_config $switches
+cp topo/common/faucet/ryu.conf $container_inst/faucet/
```
grafnu, line 90: this should just be * at the end, not ryu.conf -- reason is to make it transparent to the script since the script...

grafnu, line 90: common copy should go before the template additions.

bin/setup_flat
```diff
@@ -72,7 +86,11 @@ container=forch-faucet-$inst
...
+cp topo/common/faucet/uniform.yaml $container_inst/faucet/
 cp -r topo/$topo/forch $container_inst/
+cp topo/common/forch/forch.yaml $container_inst/forch/
```
grafnu, line 93: common copy should go before $topo version

bin/setup_flat
```diff
@@ -1,5 +1,17 @@
...
 
+function setup_flat_faucet_config {
+    local switch_count=1
+    if [ -n "$1" ]; then
+        switch_count=$1
+    fi
+    echo "Creating Faucet config for flat topology with $switch_count switches"
+    local root=$(realpath $(dirname $0)/..)
+    local out_file_arg="-p $root/inst/forch-faucet-1/faucet/faucet.yaml"
+    local topo_arg="-t flat -a $switch_count -d 1"
```
grafnu, line 11: _arg -> _args

testing/test_base
```diff
@@ -9,6 +9,7 @@ export TEST_RESULTS
 
 testing/test_check
 testing/test_topology
+testing/test_flat cleanup
```
grafnu, line 12: Best to keep the arg syntax same as the others -- so no "cleanup" arg here... Just default to cleaning. If there's a reason to not cleanup then I'd go with the inverse with "noclean" or similar.

etc/test_base.out
```diff
@@ -3,6 +3,15 @@ testing/test_base
 RADIUS request for 00:aa:bb:cc:dd:ee
 auth_callback for 00:aa:bb:cc:dd:ee: access:Authorized segment:test role:test
 0
+Testing connectivity
```
grafnu, line 6: Why is this included twice (in test_base.out and test_flat.out)?

bin/setup_flat
didovesei: Done

bin/setup_flat
didovesei: Copying the directory and using * in the command below where we copy the fiels in $topo/faucet directory

didovesei: Done

bin/setup_flat
didovesei: Done

bin/setup_flat
didovesei: Done

testing/test_base
didovesei: Ok, changed the variable to `no_clean`in testing/test_flat and removed `cleanup` here

etc/test_base.out
didovesei: This was used to run the test_flat independently. Removed now..

grafnu (approved):

PR #225, opened 2020-12-02
Batch jobs to cap max number of open processes

anurag6:


testing/python_lib/integration_base.py
```diff
@@ -155,11 +155,23 @@ def parallelize(self, target, target_args):
             process = multiprocessing.Process(target=target, args=arg_tuple)
             jobs.append(process)
 
-        for job in jobs:
-            job.start()
+        batch_start = 0
+        if not batch_size:
```
grafnu, line 159: generally best to avoid modifying a varaible, and rather create a new one, e.g. 

batch_size_actual = batch_size or len(jobs)

the reason is it makes the code easier to understand, because I can understand the beginning, and the end, without having to understand the middle.

testing/python_lib/integration_base.py
```diff
@@ -155,11 +155,23 @@ def parallelize(self, target, target_args):
...
+            batch_size = len(jobs)
+        while batch_start  <= len(jobs):
+            batch_end = batch_size + batch_start
```
grafnu, line 162: batch_start + batch_size

testing/python_lib/integration_base.py
```diff
@@ -145,7 +145,7 @@ def _get_faucet_config_path(self):
         return os.path.dirname(os.path.abspath(__file__)) + \
             (config_file_format % ('forch-faucet-1'))
 
-    def parallelize(self, target, target_args):
+    def parallelize(self, target, target_args, batch_size=None):
```
grafnu, line 148: seems like it might be better to default to 200?  Seems like the common case would be to limit things, and then if necessary people could explicitly make it bigger?

grafnu (commented):
and stickler!

testing/python_lib/integration_base.py
anurag6: Makes sense to me.

testing/python_lib/integration_base.py
anurag6: Changed.

testing/python_lib/integration_base.py
anurag6: Changed

grafnu (approved):

PR #226, opened 2020-12-02
Extend build_config script to support more topology types

didovesei:


bin/stack_functions
```diff
@@ -431,6 +431,14 @@ function reset_system {
     fi
 }
 
+function setup_corp_faucet_config {
+    echo "** setting corp config"
```
grafnu, line 435: Keep debugging echo similar style to elsewhere (i.e. nothing else is prefixed by **)

testing/python_lib/build_config.py
```diff
@@ -62,18 +124,53 @@ def _create_scale_faucet_config(self, t1_switches, t2_switches, access_ports):
...
             tap_vlan = test_vlan if not dp_index else None
-            interfaces = self._build_t1_interfaces(dp_index, t1_dps, t2_dps,
-                                                   50 + dp_index * 2, [setup_vlan], tap_vlan)
+            interfaces = self._build_dp_interfaces(
+                dp_index, dps=t1_dps, t2_dps=t2_dps, tagged_vlans=[setup_vlan],
+                tap_vlan=tap_vlan, egress_port=FAUCET_EGRESS_PORT, lacp=True)
             dps[dp_name] = self._build_datapath_config(
-                177 + dp_index, ('0e:00:00:00:01:%s' % ("{:02x}".format(dp_index+1))), interfaces)
+                T1_DP_ID_START + dp_index, interfaces,
+                (T1_DP_MAC_PREFIX + "{:02x}".format(dp_index+1)))
```
grafnu, line 132: NB: I see this bit of code 3x+ times, so is a good candidate for extracting to a function.  Right now it takes a bit of mental energy to parse what this is doing, so having a named function would clarify for the reader.

testing/python_lib/build_config.py
```diff
@@ -62,18 +124,53 @@ def _create_scale_faucet_config(self, t1_switches, t2_switches, access_ports):
...
+        return FaucetConfig(dps=dps, version=2, include=['uniform.yaml'], vlans=vlans)
+
+    def create_corp_faucet_config(self):
+        """Create Faucet config for corp network"""
+        setup_vlan = SETUP_VLAN
+        switch = 'corp'
+        dps = {}
+
+        interfaces = self._build_dp_interfaces(
+            CORP_DP_ID, tagged_vlans=[setup_vlan], access_ports=1, access_port_start=1,
```
grafnu, line 168: Shouldn't access_ports be parameterized?  How does this work for setups with 2+ egress uplinks?

grafnu (commented):
Overall looks good -- nice work!  Just a few small style things and then one "but how does this work??!?!" kind of question.

bin/stack_functions
didovesei: Actually I was going to remove but forgot.. Removed now

testing/python_lib/build_config.py
didovesei: Done

testing/python_lib/build_config.py
didovesei: So the `access_ports` specifies the number of ports that has native_vlan. The `egress_port` specifies the port id for the port with tagged_vlans.

This function will create a faucet config as:
```
dps:
  corp:
    dp_id: 273
    hardware: Generic
    interfaces:
      1:
        description: IoT Device
        max_hosts: 1
        native_vlan: 171
      10:
        description: egress
        tagged_vlans:
        - 171
......
```

`access_ports=1`  results in only one native_vlan port which is port 1, and `egress_port=CORP_EGRESS_PORT` results in port 10, which is connected to the bonded uplinks in Faucet network.

In the stack I assume we always use bonded uplinnks, which means we only need one port with tagged_vlans in corp switch..

grafnu: Ok, I was associating "access_ports" with "access_switches" which is not at all the same.  That said, my question is then how do you specify the number of south-facing ports?  Your example only shows one, but what about N (for N access switches)?

didovesei: So I am wondering if there is such a case where each south-facing port of the corp switch is connected to a Faucet switch. I thought that in our testing setup, there is always only one south-faceing port that is connected to a "bond" interface. This bond interface furuther bonds several interfaces that are connected to the egress ports of the faucet switches. That's why I assumed we only require one port with tagged vlans for corp switch.

For example, in setup_stack, we have:
```
sudo ip link add bond type bond mode 802.3ad
sudo ip link set bond up
sudo ip link set corp-eth10 master bond
sudo ip link set corp-eth20 master bond
```

And for the flat topology, the uplinks are bonded as well.

grafnu:
Oh, you're right -- I was getting the bonding setup confused with the corp
switch, my bad.

didovesei:
> Oh, you're right -- I was getting the bonding setup confused with the corp switch, my bad.

I see.. Can I go ahead merging it? I have resolved the other two comments.

grafnu (approved):

PR #227, opened 2020-12-02
Preserve directory path of included file in Faucet config

didovesei:
For example, if the included file is `foo/bar/uniform.yaml`, with this PR, Forch will write it to `/etc/faucet/foo/bar/uniform.yaml`. Currently the output path is `/etc/faucet/uniform.yaml`.

### DIFF
--- a/forch/utils.py
+++ b/forch/utils.py
@@ -93,6 +93,7 @@ def str_proto(message_text, proto_func):
 
 
 def safe_mkdir(path):
+    """Create directory if not exists"""
     try:
         os.makedirs(path)
     except OSError as error:
--- a/forch/utils.py
+++ b/forch/utils.py
@@ -1,6 +1,5 @@
 """Utility functions for forch"""
 
-import errno
 import logging
 import os
 import yaml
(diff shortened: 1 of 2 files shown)


forch/utils.py
```diff
@@ -89,3 +90,12 @@ def dict_proto(message, proto_func, ignore_unknown_fields=False):
 def str_proto(message_text, proto_func):
     """Convert a string represented protobuf message to proto object"""
     return text_format.Parse(message_text, proto_func())
+
+
+def safe_mkdir(path):
+    try:
+        os.makedirs(path)
```
grafnu, line 98: How is this functionally different than

`os.makedirs(new_path, exist_ok=True)`

didovesei: ah, I didn't know there is such a parameter.. Updated

grafnu (commented):
Lots of failing tests but the code itself looks ok...

didovesei:
> Lots of failing tests but the code itself looks ok...

I have fixed the tests. PTAL..

grafnu (approved):

PR #231, opened 2020-12-08
Use FORCH_LOG env for Forch log location

didovesei:


grafnu (approved):

didovesei:
> this is not the right behavior... if log directory does not exist, the system should 1) create it or 2) throw a fatal error.

Updated. Now if it does not exist Forch will create it. Also made some changes to make the loggers private to the classes, since when we import the module, we should not import its logger... PTAL..

grafnu (approved):

PR #232, opened 2020-12-09
Add option in forch.yaml for keepalived.pid location

didovesei:


grafnu (approved):
Code looks ok, but all sorts of failing tests...

didovesei:
> Code looks ok, but all sorts of failing tests...

I think that's because I forgot to commit the built proto changes.. Pushed the changes and the tests seem to be fine now...

PR #233, opened 2020-12-10
Add varz to expose static vlan assignment

didovesei:


forch/port_state_manager.py
```diff
@@ -107,7 +109,8 @@ class PortStateManager:
 
     # pylint: disable=too-many-arguments
     def __init__(self, process_device_placement, process_device_behavior, get_vlan_from_segment,
-                 update_device_state_varz=None, testing_segment=None):
+                 update_device_state_varz=None, update_static_vlan_varz=None,
```
grafnu, line 112: this pattern (argument list of callback functions) it's right -- it's starting to collect a bunch of things in it and it's not clean code. We should think a bit about the abstraction here and make sure we're implementing all of this correctly.  It might mean that we need another object/class that encapsulates some of this behavior and pass that one class in, rather than a bunch of separate callbacks!

Right now, there are ~5 functions here being passed in... how many different target objects do they go to?

didovesei:
Hi all, I have added two interfaces to encapsulate the callbacks passed to PortStateManager. PTAL..

grafnu (approved):
Just one comment about where the interface class files reside... but otherwise looks good!

didovesei:
> Just one comment about where the interface class files reside... but otherwise looks good!

The interface classes are in forch/device_state_manager.py and forch/varz_updater.py. I guess the commend was not added?

grafnu:
Oh, I'll have to check about the comment not being added. But it just said
"I don't think these need the be in separate modules (.py files)"... The
pythonic way to do things seems to be a few classes in some files...

grafnu:
I'd move them into the existing files. If there's some issues with import
loops we can deal with that in the future.

PR #234, opened 2020-12-10
Add lock for updating config writing time

didovesei:


forch/forchestrator.py
```diff
@@ -121,6 +121,7 @@ def __init__(self, config):
             _CONFIG_HASH_VERIFICATION_TIMEOUT_SEC_DEFAULT)
 
         self._lock = threading.Lock()
+        self._timer_lock = threading.Lock()
```
anurag6, line 124: NIT: Should the name be more specific? Like config_change_timer_lock or something?

anurag6 (approved):

grafnu:
And/or the previous "self._lock" should be renamed to something more
specific.

didovesei:
Renamed `self._lock` to `self._states_lock`. So my idea was that the each can be reused for multiple variables as long as they don't cause deadlocks. That's why I gave it a more general name..

PR #237, opened 2020-12-10
Refactor config files structure in unit tests

didovesei:
This is to fix test_base for fot branch

### DIFF
--- a/testing/python_lib/unit_base.py
+++ b/testing/python_lib/unit_base.py
@@ -31,27 +31,33 @@ class UnitTestBase(unittest.TestCase):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._temp_dir = None
-        self._temp_structural_config_file = None
-        self._temp_behavioral_config_file = None
-        self._temp_forch_config_file = None
-        self._temp_socket_file = None
+        self._base_dir = None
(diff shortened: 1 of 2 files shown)


testing/python_lib/unit_base.py
```diff
@@ -70,12 +76,12 @@ def __init__(self, *args, **kwargs):
         self._forchestrator = None
 
     def _setup_env(self):
-        assert self._temp_dir
-        os.environ['FORCH_CONFIG_DIR'] = os.path.join(self._temp_dir, 'forch')
-        os.environ['FORCH_CONFIG_FILE'] = os.path.basename(self._temp_forch_config_file)
-        os.environ['FAUCET_CONFIG_DIR'] = os.path.join(self._temp_dir, 'faucet')
-        os.environ['FAUCET_CONFIG_FILE'] = os.path.basename(self._temp_behavioral_config_file)
-        os.environ['FAUCET_EVENT_SOCK'] = self._temp_socket_file
+        assert self._base_dir
```
grafnu, line 79: why assert _base_dir here?  It's not used in this function...

didovesei: Right, it's not used below in that function. Removed. PTAL..

anurag6 (approved):

PR #239, opened 2020-12-11
Set static ip assignment in DHCP server for faux-1 and faux-2

didovesei:


anurag6:
Would this need a change similar to what we do with faux-3? https://github.com/faucetsdn/forch/blob/master/bin/setup_stack#L145

didovesei:
> Would this need a change similar to what we do with faux-3? https://github.com/faucetsdn/forch/blob/master/bin/setup_stack#L145

Yes, it is already in the target branch. Basically in that branch we are doing dhcp for faux-1 and faux-2 instead of faux-3

anurag6 (approved):

PR #245, opened 2021-01-06
Feature/port up down events

henry54809:


testing/python_test
```diff
@@ -13,12 +13,14 @@ if [ "$1" == coverage ]; then
...
-    echo $0 [test_name]
+if [ $# == 0 ]; then
+    echo $0 test_name [class_name]
+    echo class_name should be the unit test class with an optional method name e.g. FotFaucetizerTestCase.test_device_states 
     false
 fi
 
 TEST_NAME=$1
+CLASS_NAME=$2
 shift
```
grafnu, line 24: shift should be conditional in this case since class is optional

testing/python_test
```diff
@@ -13,12 +13,14 @@ if [ "$1" == coverage ]; then
     shift
 fi
 
-if [ $# != 1 ]; then
-    echo $0 [test_name]
+if [ $# == 0 ]; then
```
grafnu, line 16: >2 arguments is also not allowed

forch/device_report_server.py
```diff
@@ -44,12 +74,20 @@ def __init__(self, on_receiving_result, address=None, port=None, max_workers=Non
         self._server = grpc.server(
             futures.ThreadPoolExecutor(max_workers=max_workers or MAX_WORKERS_DEFAULT))
 
-        servicer = DeviceReportServicer(on_receiving_result)
-        device_report_pb2_grpc.add_DeviceReportServicer_to_server(servicer, self._server)
+        self._servicer = DeviceReportServicer(on_receiving_result, )
```
grafnu, line 77: I'm curious what the comma does... isn't that a syntax error?

forch/forchestrator.py
```diff
@@ -418,7 +418,22 @@ def _register_handlers(self):
                 event.timestamp, event.dp_name, event.port_no, event.eth_src, event.l3_src_ip)),
             (FaucetEvent.L2Expire, lambda event: fcoll.process_port_expire(
                 event.timestamp, event.dp_name, event.port_no, event.eth_src, event.vid)),
-        ])
+        ]
+        if self._device_report_server:
+            handlers.extend([
+                (
+                    FaucetEvent.PortChange,
+                    lambda event: self._device_report_server.process_port_change(
```
grafnu, line 426: don't include this logic here -- keep it super simple. So, the handlers being added should be unconditional, and then any logic (conditional) should be handled in a sub method.  Yes, it adds a few (two?) helper methods, but the goal/intent is to keep the dispatch table very clean/simple.

forch/device_report_server.py
```diff
@@ -21,6 +23,22 @@ def __init__(self, on_receiving_result):
         super().__init__()
         self._on_receiving_result = on_receiving_result
         self._logger = get_logger('drserver')
+        self._port_device_mapping = {}
+        self._port_events_listeners: Dict[str, List[Queue]] = {}
```
grafnu, line 27: stickler error... not even sure myself what the line is supposed to be doing!

henry54809: I tried to put a type definition there but that's generating a false positive from flake8.

forch/device_report_server.py
henry54809: I removed the comma but that's not a syntax error in python; the same way a tuple can have an extra comma in the end e.g. (1,2,)

forch/forchestrator.py
henry54809: I updated. PTAL and see if my changes match what you had in mind.

grafnu: Not quite... What I was going for was basically that the dispatch table was static, and *always* had a port change event handler... and then the conditional logic was inside of the port change handler, i.e.:

`(FaucetEvent.PortChange, lambda event: self._process_port_chance(event))`

and then

```
def _process_port_change(self, event):
  if self._device_report_server:
    self.device_report_server.process_port_change(event)
```

or... create a "null" device_report_server that is instantiated if there's not a full one used. So, the method is always called, but sometimes it just does nothing.

henry54809: faucet events class was updated to accept multiple handlers for each event, so in this case FaucetEvent L2Learn and PortChange have handlers both from the device report server and the faucet state collector. Using your suggestion and have the multiple handlers support are going to be counter intuitive. what do you think?

grafnu:
I think they're orthogonal considerations. With my suggestion, there's
nothing that prevents it from working and still being "clean" in the sense
that the list (rather than map) of event handlers is still static, not
conditional. My experience has been that it's generally just confusing when
"sometimes a handler is there and sometimes it's not" -- and a lot easier
to reason about it when the existence of the handler itself is a constant,
but what the handler does is where the variance occurs. It's just moving
the can (conditional logic) around, but if the handler list itself is
dynamic then it effectively gets pushed down into the event dispatcher
which makes it a lot harder to debug down the road. (And yes, I understand
that the actual conditional is still contained in the class itself, but the
ultimate decision point is then in the dispatch logic, which is where the
confusion arises).  Having multiple handlers for a given event is
pretty common, so I don't think it's counter intuitive at all.

henry54809:
PTAL

forch/forchestrator.py
```diff
@@ -414,32 +414,26 @@ def _register_handlers(self):
...
-                    event.timestamp, event.dp_name, event.port_no,
-                    event.status and event.reason != 'DELETE')
-            ),
-            (
-                FaucetEvent.L2Learn,
-                lambda event: self._device_report_server.process_port_learn(
-                    event.dp_name, event.port_no, event.eth_src)
-            ),
-        ]
+    def _device_report_server_process_port_change(self, event):
```
grafnu, line 427: nit: personally I don't believe it's good to over-specify the function name, since really it could just be "process_port_change" -- since if, for some reason there's another thing you want to do on a port change then you could also use this function.  Really there's no need to have two entries in the dispatch table (you could move the fcoll.process_port_change call into this function).  Doesn't really matter -- thus the nit.  Change or not as you see fit.

grafnu (approved):

PR #255, opened 2021-01-19
Make one device pass and one device fail in FOT test

didovesei:


controller/bin/daq_system_default.yaml
```diff
@@ -0,0 +1,51 @@
+---
```
grafnu, line 1: why pull in a copy of this file?

controller/bin/daq_config.yaml
```diff
@@ -1,6 +1,6 @@
 # DAQ FOT configuration.
 
-include: ../daq/config/system/default.yaml
+include: daq_system_default.yaml
```
grafnu, line 3: why create a new separate file?

bin/setup_stack
```diff
@@ -178,7 +184,7 @@ fi
 
 echo Connecting faux interfaces...
 add_faux t2sw1 1 1 $faux_dhcp
-add_faux t2sw2 1 2 $faux_dhcp
+add_faux t2sw2 1 2 $faux_dhcp $faux_http
```
grafnu, line 187: can this just be fixed?  Does anything bad happen if we just always enable http on this device?  I'm trying to keep the overall complexity down unless there's a specific reason to make it configurable!

etc/DAQ_VERSION
```diff
@@ -1 +1 @@
-1.9.13
+master
```
grafnu, line 1: Is there a specific DAQ version/change you need?  It's trivial for us now to tag an intermediate DAQ version (e.g. 1.9.14) -- so we can just do that if necessary.

etc/start_faux
```diff
@@ -53,5 +53,10 @@ if [ -n "${options[dnsmasq]}" ]; then
     dnsmasq --log-facility=/tmp/dnsmasq.log --log-dhcp
 fi
 
+if [ -n "${options[http]}" ]; then
+    echo Listening on TCP 80...
+    python3 -m http.server &
```
henry54809, line 58: This by default listens on port 8000.

controller/bin/daq_config.yaml
didovesei: Reverted. I had some path issue with some DAQ version but now it is working. Not sure what fixed it...

bin/setup_stack
didovesei: Done. The http server is always running on faux-2 now.

etc/DAQ_VERSION
didovesei: Using 1.9.14 now

etc/start_faux
didovesei: Updated the echo message. I saw daq does fail the device even if it is listing on 8000.

controller/bin/daq_system_default.yaml
didovesei: Removed

grafnu (approved):

PR #258, opened 2021-01-22
Cleanup stale device policies

didovesei:


forch/port_state_manager.py
```diff
@@ -143,38 +143,54 @@ def handle_device_behavior(self, mac, device_behavior, static=False):
...
+                self._set_port_sequestered, self._set_port_operational,
+                self._handle_infracted_state)
+
+            device_behavior = (self._static_device_behaviors.get(mac) or
+                               self._dynamic_device_behaviors.get(mac))
+            if device_behavior:
+                static = mac in self._static_device_behaviors
+                self.handle_device_behavior(mac, device_behavior, static=static)
+
+        return True, None, stale_mac
```
grafnu, line 178: note this is fine for now, but it's headed in a problematic direction -- in the long run, it makes things much harder to understand with complex tuple return values, esp. when the point of call just returns that again.  Since python (by default, how we use it) doesn't specify return types in the method signature, it means a lot of spaghetti knowledge in the code to figure things out.

Also, it's losing the parity with the function title ("handle") -- since it's doing a lot more than handle the event.

Anyway, fine for now, but next time you find yourself trying to overload a tuple return value, consider ways to restructure the code so that it doesn't need to work that way (there's a whole bunch of ways to accomplish that, just depending on the situation).

grafnu (approved):

PR #262, opened 2021-01-25
Send traffic to faux-1 as well to make tests more deterministic

anurag6:


grafnu (approved):

didovesei (approved):

1
```diff
@@ -0,0 +1,178 @@
+#!/bin/bash
```
didovesei, line 1: Seems that there is a duplicate file?

anurag6: Yeah, no I committed in debug files by mistake. Was just reverting them.

PR #268, opened 2021-01-28
Make segments_to_vlans and tail_acl mandatory for running DVA

didovesei:


forch/forchestrator.py
```diff
@@ -297,22 +299,33 @@ def _calculate_config_files(self):
...
-        return False
+        if not os.path.exists(self._structural_config_file):
+            raise Exception(
+                f'Structural config file does not exist: {self._structural_config_file}')
+
+        self._segments_vlans_file = os.path.join(
+            self._forch_config_dir, orch_config.segments_vlans_file)
+        if not os.path.exists(self._segments_vlans_file):
+            error_msg = f'Segments-to-vlans file does not exist: {self._segments_vlans_file}.'
+            error_msg += ' DVA disabled'
```
grafnu, line 317: would prefer a single assignment rather than appending a string.  You can do this with % notation and ()'s or something if the line gets too long.

didovesei: updated. PTAL

grafnu (approved):

PR #273, opened 2021-02-03
Refactoring and cleanup

grafnu:


bin/check_style
```diff
@@ -32,10 +32,17 @@ failure=
 results=
 count=0
 for file in forch/*.py testing/python_lib/*.py; do
-    echo Checking file $file... | tee -a $RESULTS
-    pylint --rcfile=$PYLINTRC $file >> $RESULTS && count=$((count+1)) || true
+    echo Checking file $file... | tee -a $RESULTS.$count
+    pylint --rcfile=$PYLINTRC $file >> $RESULTS.$count &
```
henry54809, line 36: this can be just > instead of >>
in case those files were removed from a previous unfinished run.

grafnu: Updated. PTAL. Tests are failing but I think that's a not-this-PR problem (normal build is failing too from what I can tell).

henry54809 (approved):

didovesei (approved):

anurag6 (approved):

PR #276, opened 2021-02-09
Fix the order of links in link graph before comparing with previous link graph

anurag6:


didovesei (approved):

grafnu (approved):
Thanks for fixing this! Such a super simple fix, but finding the problem is never fun...

PR #277, opened 2021-02-11
Add vlan packet count for each port and host in NOAH API

didovesei:


proto/vlan_state.proto
```diff
@@ -0,0 +1,12 @@
+/*
+ * Port VLAN state
+ */
+syntax = "proto3";
+
+message VLANState {
```
grafnu, line 6: Should be VlanState (in this context, 'vlan' is considered a word, so it would be VlanState or vlan_state depending on the language style).

didovesei: Updated. PTAL..

proto/vlan_state.proto
```diff
@@ -3,7 +3,7 @@
  */
 syntax = "proto3";
 
-message VLANState {
```
grafnu: How is this reconciled with the VLANState that's in switch_state.proto?  Seems odd to have two of the same thing floating around.  Also, this seems too small to have it's own proto file -- what's the higher-level concept that this fits into?  Should it just go in shared constants?  Proto files shouldn't necessarily follow the one-file-per-message pattern, since that just ends up with a lot of really small files for no good reason.  Seems like it's conceptually similar to DVAState so I would collocate with that.

didovesei: > How is this reconciled with the VLANState that's in switch_state.proto? Seems odd to have two of the same thing floating around
The old `VLANState` in the switch_state is supposed to be used here for the port vlan state. However, the problem is that it is calculated based on the `flow_packet_count_vlan_acl` varz that only exposes the vlan metrics at the switch level. There is no "in_port" field in that varz's labels.

The new `VlanState` is calculated from `flow_packet_count_vlan` varz that has per port vlan metrics, but this varz does not have the vlan ACL info. Therefore there are two different protos with totally different structures created for the same thing.

I renamed old `VLANState` to `VlanAclstate` since it mainly exposes the ACL states for each switch.

> what's the higher-level concept that this fits into?
I moved the AclState, VlanAclState and VlanState into the new file network_metric_state.proto.

grafnu (approved):
Very nice -- thanks for taking the time to think through the proper organization!

PR #280, opened 2021-02-18
Refactor DeviceReportServer methods with interface

didovesei:
* Added interface `DeviceStateReporter`
* Make Forch only report applied vlan to DAQ, when device has both static and dynamic vlan assignment


grafnu (approved):
Nice, I like it -- that's a good refactoring!

PR #286, opened 2021-03-01
Add vrrp_state API

didovesei:


forch/http_server.py
```diff
@@ -56,6 +57,7 @@ def do_GET(self):
             self.send_response(http.HTTPStatus.INTERNAL_SERVER_ERROR)
             self.end_headers()
             self._logger.error('Unhandled exception: %s', exception)
+            traceback.print_tb(exception.__traceback__)
```
grafnu, line 60: use logging.exception() for this

forch/__main__.py
```diff
@@ -59,6 +59,7 @@ def run_forchestrator():
         http_server.map_request('process_state', forchestrator.get_process_state)
         http_server.map_request('host_path', forchestrator.get_host_path)
         http_server.map_request('list_hosts', forchestrator.get_list_hosts)
+        http_server.map_request('vrrp_state', forchestrator.get_vrrp_state)
```
grafnu, line 62: I think this is too small to be its own top-level API... it should be part of system_state  (unless there's a really good reason it needs to be separate).

forch/forchestrator.py
```diff
@@ -739,6 +739,9 @@ def _get_controller_state(self):
             active_state = self._active_state
             if active_state == State.initializing:
                 return State.initializing, 'Initializing'
+            if active_state == State.inactive:
```
grafnu, line 742: else if

forch/http_server.py
didovesei: Done

forch/forchestrator.py

forch/__main__.py
didovesei: So this was actually Eng's request. They would like VRRP state to be a subsystem like the others, and to display the detailed vrrp error through vrrp_state API instead of the system_state_detail.

grafnu: Ok -- just something to be careful of (in terms of breaking encapsulations), but sounds like a reasonable enough ask!

grafnu (approved):

forch/forchestrator.py
didovesei: Actually this else if not needed, right? Indeed, there seems to be a lint error about this:
```
forch/forchestrator.py:740: [R1705(no-else-return), Forchestrator._get_controller_state] Unnecessary "elif" after "return"
```

grafnu:
Oh, I suppose you're right given the return statement!

anurag6 (approved):

PR #287, opened 2021-03-01
Validate tail_acl and segments_vlans_file during Forch initialization

didovesei:
The main idea of this change is that if either of these two configs are invalid (specifed acl/file does not existing, file reading error, etc.), Forch will continue to run with static and dynamic auth disabled. Forch will also display the detailed error on NOAH.

### DIFF
--- a/forch/faucetizer.py
+++ b/forch/faucetizer.py
@@ -122,9 +122,9 @@ class Faucetizer(DeviceStateManager):
 
             self.flush_behavioral_config()
 
-    def validate_tail_acl_config(self, tail_acl):
+    def validate_tail_acl_config(self):
         """Validate tail_acl config"""
-        return self._has_acl(tail_acl)
+        return not self._config.tail_acl or self._has_acl(self._config.tail_acl)
 
     def _process_structural_config(self, faucet_config):
         """Process faucet config when structural faucet config changes"""
@@ -225,8 +225,6 @@ class Faucetizer(DeviceStateManager):
     def _finalize_host_ports_config(self, behavioral_faucet_config, new_testing_device_vlans):
         testing_port_vlans = new_testing_device_vlans.values()
         testing_port_configured = False
-        tail_acl = self._config.tail_acl
-        has_tail_acl = tail_acl and self.validate_tail_acl_config(tail_acl)
 
         for switch_map in behavioral_faucet_config.get('dps', {}).values():
             for port_map in switch_map.get('interfaces', {}).values():
@@ -234,7 +232,8 @@ class Faucetizer(DeviceStateManager):
                 if port_type == PortType.testing and testing_port_vlans:
                     port_map.setdefault('tagged_vlans', []).extend(testing_port_vlans)
                     testing_port_configured = True
-                if self._get_port_type(port_map) == PortType.access and has_tail_acl:
+                if (self._get_port_type(port_map) == PortType.access and
+                        self.validate_tail_acl_config()):
                     port_map.setdefault('acls_in', []).append(self._config.tail_acl)
 
         if testing_port_vlans and not testing_port_configured:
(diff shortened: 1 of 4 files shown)


forch/forchestrator.py
```diff
@@ -220,10 +221,27 @@ def _initialize_orchestration(self):
             self._initialize_faucetizer(sequester_segment)
             self._faucetizer.reload_structural_config()
 
+            tail_acl = self._config.orchestration.tail_acl
+            if tail_acl and not self._faucetizer.validate_tail_acl_config(tail_acl):
+                error_msg = 'DVA auth was disabled due to missing ACL for tail_acl config'
```
grafnu, line 226: We need to be consistent here.  Is "DVA" an actual feature of Forch or is it a collection of features?  If it's a collection, then this error message should be explicit about what is actually disabled.

forch/faucetizer.py
```diff
@@ -225,13 +225,15 @@ def _initialize_host_ports(self, behavioral_faucet_config):
     def _finalize_host_ports_config(self, behavioral_faucet_config, new_testing_device_vlans):
         testing_port_vlans = new_testing_device_vlans.values()
         testing_port_configured = False
+        has_tail_acl = self._config.tail_acl and self._has_acl(self._config.tail_acl)
```
grafnu, line 228: THis doesn't seem consistent with the validate function you created.  Why not just use the validate function here?  Should the check for self._config.tail_acl also be in the validate function?

grafnu (commented):
bit confused because I thought we were going with the "if a tail_acl is not configured then it will use a default "empty" acl instead" to avoid the reconfiguration problem... ?  Either way, just checking that this WAI and does what is needed.

didovesei:
> bit confused because I thought we were going with the "if a tail_acl is not configured then it will use a default "empty" acl instead" to avoid the reconfiguration problem... ? Either way, just checking that this WAI and does what is needed.

So after having discussed with Eng, we agreed that the existence of tail_acl configuration will be validated during config generation. Forch only needs to handle the case where the tail_acl is configured but no corresponding ACL is present.

grafnu:
Got it, sounds good, thanks.

forch/forchestrator.py
didovesei: Update the message to "All auth was disabled...".

forch/faucetizer.py

forch/faucetizer.py
```diff
@@ -122,9 +122,9 @@ def process_device_behavior(self, eth_src, behavior, static=False):
 
             self.flush_behavioral_config()
 
-    def validate_tail_acl_config(self, tail_acl):
+    def validate_tail_acl_config(self):
```
grafnu, line 125: validate is a verb but this function doesn't do anything.  tail_acl_config_valid()?

forch/faucetizer.py
```diff
@@ -225,13 +225,15 @@ def _initialize_host_ports(self, behavioral_faucet_config):
     def _finalize_host_ports_config(self, behavioral_faucet_config, new_testing_device_vlans):
         testing_port_vlans = new_testing_device_vlans.values()
         testing_port_configured = False
+        should_apply_tail_acl = self._config.tail_acl and self.validate_tail_acl_config()
```
anurag6, line 228: Correct me if I have this wrong:
A = self._config.tail_acl; B = self._has_acl(self._config.tail_acl); A' = not self._config.tail_acl;

validate_tail_acl_config = A' or B

should_apply_tail_acl = A and (A' or B)
                                      = (A and A') or (A and B)
                                      = None or (A and B).  = A and B

Logically, would it be the same thing to simplify should_apply_tail_acl = self._config.tail_acl and self._has_acl(self._config.tail_acl)

anurag6, line 228: NIT: I think 'should_' is redundant

anurag6: My bad. Just checked comments above.

anurag6 (approved):

forch/faucetizer.py
didovesei: right.. Done

forch/faucetizer.py
didovesei: I see..

grafnu (approved):

PR #290, opened 2021-03-07
Always use lowercase MAC in PortStateManager

didovesei:
There was an issue where static device behavior was defined with uppercase MAC but the corresponding dynamic device placement was learned with lowercase MAC. As a result PortStatemanager treated them as different devices.

### DIFF
--- a/forch/forchestrator.py
+++ b/forch/forchestrator.py
@@ -817,7 +817,7 @@ class Forchestrator(VarzUpdater, OrchestrationManager):
                 is_stack = 1 if 'stack' in if_obj else 0
                 is_access = 1 if 'native_vlan' in if_obj else 0
                 is_tap = 1 if if_obj['description'] == 'tap' else 0
-                is_mirror = 1 if if_obj['description'].lower() == 'mirror' else 0
+                is_mirror = 1 if if_obj['description'] == 'MIRROR' else 0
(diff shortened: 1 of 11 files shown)


forch/forchestrator.py
```diff
@@ -816,7 +817,7 @@ def _validate_config(self, config):
                 is_stack = 1 if 'stack' in if_obj else 0
                 is_access = 1 if 'native_vlan' in if_obj else 0
                 is_tap = 1 if if_obj['description'] == 'tap' else 0
-                is_mirror = 1 if if_obj['description'] == 'mirror' else 0
+                is_mirror = 1 if if_obj['description'].lower() == 'mirror' else 0
```
grafnu, line 820: is this strictly necessary?  Generally it's better to be more exact in matching, not permissive. Why is there a need for this to handle both MIRROR, mirror, MiRRor, etc... ?

didovesei: Updated to "MIRROR". That's what are using in productin...

didovesei: Btw, should we change "tap" to "TAP' to match the style in production? There we always uppercase prefixes for all the port descriptions.

grafnu:
Yes, that would be good (TAP vs. tap)

didovesei:
> Yes, that would be good (TAP vs. tap)

Updated. PTAL..

PR #291, opened 2021-03-07
Do not deauthenticate device after previous successful auth times out

didovesei:
Changing the logic to:
* If device is in request state and the radius request times out, device will be deauthenticated.
* If device is in authenticated state and the authentication times out, Forch reauthenticates the device without deauthenticating it firstly.
* If device is in unauthenticated state, Forch does not deauthenticate it again, since device has already been deauthenticated previously.

Note: here "deauthenitcating a device" means that Forch sends a UNAUTH notification to PortStateManager to remove the the device behavior and the port will be moved to unauthenticated VLAN.


grafnu:
Can you maybe add a comment block at the top of the
simple_auth_state_machine.py file that shows all the states and a rough one
sentence description of what it means, and then relevant comments about the
transitions out?  The words you wrote here are good, but they would be
better captured in the file itself.

I'm not convinced the "request" state logic is correct, but it's hard for
me to really say for sure without a more clear detailing of what the state
is supposed to mean.

It seems there's two different parameters to the state machine -- the
authenticated state, and the request state -- and I want to make sure
they're captured correctly and transitions are logical.

The one thing that might be getting lost here is the notion of "retries"
for an authentication timeout... usually one timeout is not sufficient to
mean "failed", and so we need to make sure that's accounted for correctly.
I think in the "simple" case it's ok since if you're not yet authenticated
then a timeout is equivalent to failure and try again, but that's not true
from a previously authenticated state.

didovesei:
> Can you maybe add a comment block at the top of the simple_auth_state_machine.py file that shows all the states and a rough one sentence description of what it means, and then relevant comments about the transitions out? 

Added comments for the states and timeout handling. 

> The one thing that might be getting lost here is the notion of "retries" for an authentication timeout... usually one timeout is not sufficient to mean "failed", and so we need to make sure that's accounted for correctly. I think in the "simple" case it's ok since if you're not yet authenticated then a timeout is equivalent to failure and try again, but that's not true from a previously authenticated state.

I see, makes sense to me. So I added a check for `self._retry_backoff`. If it is set, which means a previous request has been sent but the response is still not received, then Forch will deauthenticate the device. If it is not set, which means we are currently sending the first request, then Forch does not deauthenticate the device.

forch/simple_auth_state_machine.py
```diff
@@ -101,7 +109,15 @@ def received_radius_reject(self):
             self._auth_callback(self.src_mac, self.UNAUTH, None, None)
 
     def handle_sm_timer(self):
-        """Handle timer timeout and check.trigger timeout behavior of states"""
+        """
+        Handle timer timeout and check.trigger timeout behavior of states:
+        * If the device is in request state and the RADIUS request times out, device will be
```
grafnu, line 114: consolidate each to one line for readability -- more is not always better. E.g.:
* REQUEST & timeout --> UNAUTH
* UNAUTH --> no change

forch/simple_auth_state_machine.py
```diff
@@ -113,10 +129,16 @@ def handle_sm_timer(self):
                 backoff_time = backoff * self._query_timeout_sec
                 self._current_timeout = time.time() + backoff_time
                 if self._current_state == self.REQUEST:
+                    if self._retry_backoff:
```
grafnu, line 132: I think we need something more explicit here (since this is about security).  From this implementation, it's somewhat obtuse when it will transition to the UNAUTH state.  Looking at this code I can't easily determine what the algorithm is.   Also, it's not clear how this maps to the comments above.  Do you have a clear description of what the behavior *should* be (independent of the implementation)?  AUTH -> AUTH and UNAUTH -> UNAUTH, and UNAUTH -> AUTH are all clear and obvious, but the AUTH -> UNAUTH process is unclear (where in the comment does it describe this?)

didovesei: Updated the comments.

For AUTH -> UNAUTH transition, it happens in `host_expired` when a device is disconnected from the switch. It will not be triggered by any timeout.

grafnu: That's not right -- eventually the device needs to be unauthenticated if it doesn't receive a valid radius response.

didovesei: Updated the timeout handling logics and added "REQUEST -> UNAUTH' transition. If the Forch does not receive an accepted radius response after `max_backoff` retries, the device will be set to UNAUTH state and will be deauthenticated. Once a device enters UNAUTH state, Forch will try to reauthenticate it again after `_rej_timeout_sec`.

forch/simple_auth_state_machine.py
```diff
@@ -95,28 +103,46 @@ def received_radius_reject(self):
...
-                        'Retrying RADIUS request for src_mac %s. Retry #%s', self.src_mac,
-                        self._retry_backoff)
-                self._radius_query_callback(self.src_mac, self.port_id)
-                backoff = min(self._retry_backoff, self._max_radius_backoff)
-                backoff_time = backoff * self._query_timeout_sec
-                self._current_timeout = time.time() + backoff_time
                 if self._current_state == self.REQUEST:
+                    if self._retry_backoff:
+                        if self._retry_backoff < self._max_radius_backoff:
+                            self._increment_retries()
```
grafnu, line 123: rename function to "_increment_retry_backoff" so it's very clear it's the same as the variable _retry_backoff

forch/simple_auth_state_machine.py
```diff
@@ -95,28 +103,46 @@ def received_radius_reject(self):
...
+                            self._resend_radius_request()
+                        else:
+                            self._reset_state_machine()
+                            self._auth_callback(self.src_mac, self.UNAUTH, None, None)
                     if self._metrics:
                         self._metrics.inc_var('radius_query_timeouts')
-                    self._increment_retries()
-                    self._logger.debug('RADIUS request timed out for %s', self.src_mac)
-                else:
+                    self._logger.error('RADIUS request timed out for %s', self.src_mac)
```
grafnu, line 130: at log at beginning of code block not end

forch/simple_auth_state_machine.py
```diff
@@ -95,28 +103,46 @@ def received_radius_reject(self):
...
-            self._current_timeout = time.time() + self._rej_timeout_sec
-            self._retry_backoff = 0
+            self._reset_state_machine()
             self._auth_callback(self.src_mac, self.UNAUTH, None, None)
 
     def handle_sm_timer(self):
-        """Handle timer timeout and check.trigger timeout behavior of states"""
+        """
+        Handle timer timeout and check.trigger timeout behavior of states:
+        * REQUEST state --request timeout & backoff < max_backoff--> REQUEST state + send request
```
grafnu, line 112: This syntax is very confusing. what's "check.trigger"? What's the -- before request mean?   Please try to clean it up  a bit so it's easier to understand.  "state" is redundant with ALLCAPS names.

forch/simple_auth_state_machine.py
```diff
@@ -95,28 +103,46 @@ def received_radius_reject(self):
...
-                    self._logger.debug(
-                        'Retrying RADIUS request for src_mac %s. Retry #%s', self.src_mac,
-                        self._retry_backoff)
-                self._radius_query_callback(self.src_mac, self.port_id)
-                backoff = min(self._retry_backoff, self._max_radius_backoff)
-                backoff_time = backoff * self._query_timeout_sec
-                self._current_timeout = time.time() + backoff_time
                 if self._current_state == self.REQUEST:
+                    if self._retry_backoff:
+                        if self._retry_backoff < self._max_radius_backoff:
```
grafnu, line 122: Something is getting confused here because "backoff" is not the same as a "retry" -- which is this supposed to be?  Please take a pass through and make sure that "backoff" is used to mean a time-based backoff, while retry refers to a integral counter.  I think the name "retry_backoff" is conflating the two together. You don't "retry backoff" -- you count the number of retries, each of which has a backoff time.  They are separable concepts (you can have backoff without retries, and you can have retries without backoff), so they shouldn't be directly mixed.

grafnu (commented):
I think the code/algorithm looks ok, but the variable/method names need some cleanup.

forch/simple_auth_state_machine.py
didovesei: renamed to `_increment_radius_retries`

forch/simple_auth_state_machine.py
didovesei: Done

forch/simple_auth_state_machine.py
didovesei: updated

forch/simple_auth_state_machine.py
didovesei: Updated all the relevant variables so that "retry" refers to the count and "backoff" refers to the seconds between two requests. PTAL..

proto/forch_configuration.proto
```diff
@@ -103,7 +103,7 @@ message OrchestrationConfig {
     int32 heartbeat_sec = 2;
 
     // Max Radius backoff count
-    int32 max_radius_backoff = 3;
+    int32 max_radius_retries = 3;
```
grafnu, line 106: update comment

grafnu (approved):

PR #292, opened 2021-03-23
Static ip device in fot

henry54809:


didovesei (approved):

etc/start_faux
```diff
@@ -40,6 +40,7 @@ else
 
     ip route add $IPADDR/16 via $GATEWAY dev $IFACE
     ip route add default via $GATEWAY
+    ping $GATEWAY &
```
grafnu, line 43: I'm not sure this is a good idea since it will horribly pollute the logs, no?

henry54809: Changed this so that static ip device only ping during fot tests.

henry54809:
@grafnu tests passed. PTAL

etc/Dockerfile.controller
```diff
@@ -36,6 +36,7 @@ RUN bin/install_forch && rm -rf forch/.git
 COPY daq/ daq/
 RUN daq/bin/setup_dev simple && rm -rf daq/.git
 
+COPY controller/daq_site_config /site
```
grafnu, line 39: blank line after

controller/daq_site_config/mac_addrs/9a02571e8f03/device_config.json
```diff
@@ -0,0 +1,3 @@
+{
```
grafnu, line 1: I would just call this controller/site_config (no daq_ prefix) -- the contents here are really about what goes on the controller machine, and it potentially has more uses outside of DAQ (e.g. used by cloud configuration tools).

grafnu (approved):

PR #293, opened 2021-03-25
Expose system authentication state via Varz and NOAH

didovesei:
The varz added will look like: 
```
# HELP system_initialization Initialization state of the system
# TYPE system_initialization gauge
system_initialization{authentication="3"} 1.0
```

A new state `authentication_state` is added in the system_state API:

```
  "system_state": "split",
  "system_state_detail": "Lost reachability to peer controller.",
  "system_state_change_count": 31,
  "system_state_last_change": "2021-03-25T00:27:29.060922",
  "system_state_last_update": "2021-03-25T00:29:10.956145",
  "authentication_state": "dynamic_only",       <---------------------------
```


proto/system_state.proto
```diff
@@ -36,11 +36,14 @@ message SystemState {
   // Last state update time
   string system_state_last_update = 9;
 
+  // Authentication state
+  SysAuthState.State authentication_state = 10;
```
grafnu, line 40: I would call this a Mode not a State.  It's not something that changes dynamically, rather it's based on the way things are configured. So, I'd replace the uses of "authentication_state" with "authentication_mode".  Does that make sense?

forch/forch_metrics.py
```diff
@@ -110,6 +110,10 @@ def _add_vars(self):
             'static_mac_vlan', 'MAC addresses with statically assigned vlan', Gauge,
             labels=['mac'])
 
+        self._add_var(
+            'system_initialization', 'Initialization state of the system', Gauge,
+            labels=['authentication'])
```
grafnu, line 115: Instead of "authentication" I'd name/label this "auth_mode"

proto/system_state.proto
didovesei: Right, updated it to `authentication_mode`.

forch/forch_metrics.py
didovesei: Update. PTAL..

grafnu (approved):

anurag6 (approved):

PR #294, opened 2021-03-26
Update fot testing to use native faucet

grafnu:


controller/bin/daq_config.yaml
```diff
@@ -1,11 +1,13 @@
 # DAQ FOT configuration.
 
-include: ../daq/config/system/default.yaml
+include: ${DAQ_LIB}/config/system/default.yaml
 
 switch_setup:
   data_intf: data0
   lo_port: 3984
+  varz_port: 5678
```
didovesei, line 8: I guess this needs to be updated to `alt_varz_port`..

didovesei (approved):

testing/test_fot
```diff
@@ -40,22 +34,21 @@ fgrep "System port 1 on dpid 1 is active True" $DAQBASE/inst/cmdrun.log
...
 
-sleep 20
-
 # Forch startup checks
 # Check Forch has successfuly started
 fgrep -o "Starting Forchestrator" $INST_BASE/forch.log | tee -a $TEST_RESULTS
 
+# Waiting for DAQ test to complete
+monitor_daq_log "Setting port set 1 to vlan 272"
+monitor_daq_log "Setting port set 2 to vlan 273"
```
henry54809, line 43: Don't these mean the start of DAQ tests?

grafnu:
Not quite -- the varz_port is for the prometheus varz, the "alt_" ports are
for the alternate server that is used to emulate a faucet stack outside of
DAQ. Basically, the alt_ prefixed ones are for test-only setups.

grafnu:
It would mean the start of the tests, but testing isn't enabled so it's
really just "things are ready to be tested" -- turns out DAQ wasn't
actually running tests before for a variety of reasons, so I added another
bug to fix that.

PR #296, opened 2021-03-29
Adding sequester timeout.

henry54809:
Still need to make changes to fot test when other PRs are resolved.


forch/forchestrator.py
```diff
@@ -72,7 +72,7 @@
 SEGMENTS_VLANS_FILE = 'segments_vlans_file'
 TAIL_ACL_CONFIG = 'tail_acl_config'
 SEQUESTER_SEGMENT_DEFAULT = 'SEQUESTER'
-
+SEQUESTER_TIMEOUT_DEFAULT = 20 * 60
```
grafnu, line 75: suffix with _SEC?

forch/forchestrator.py
```diff
@@ -247,7 +247,8 @@ def _initialize_orchestration(self):
             self._faucet_collector.set_device_state_reporter(self._device_report_server)
 
         self._port_state_manager = PortStateManager(
-            self._faucetizer, self, self._device_report_server, sequester_segment)
+            self._faucetizer, self, self._device_report_server,
+            sequester_segment=sequester_segment, sequester_timeout=sequester_timeout)
```
grafnu, line 251: is now the time to create a generic "config" bucket to pass in... or leave it as is? Not sure myself -- just asking the question. This is how constructors with 20 arguments start!

henry54809:
PTAL. 
I don't think it's quite necessary to create a separate sequester config object until we need a 3rd parameter other than sequester_segment and sequester_timeout.

controller/bin/daq_config.yaml
```diff
@@ -23,3 +23,5 @@ external_subnets:
   - subnet: 192.168.1.0/24
 
 site_path: /site
+
```
grafnu, line 26: mmmm... this change doesn't seem right. -- I was hoping to keep this file as somewhat of an example of how FOT should be configured, and this starts to break that.  Can this option be moved to a command-line flag for the specific test run where it's required?  Or, if specifically for this timeout test case, invert the "enabled" use below so you only specifically enable it for the DUT you want to timeout?

forch/forchestrator.py
```diff
@@ -72,7 +72,7 @@
 SEGMENTS_VLANS_FILE = 'segments_vlans_file'
 TAIL_ACL_CONFIG = 'tail_acl_config'
 SEQUESTER_SEGMENT_DEFAULT = 'SEQUESTER'
-SEQUESTER_TIMEOUT_DEFAULT = 20 * 60
+SEQUESTER_TIMEOUT_SEC_DEFAULT = 20 * 60
```
grafnu, line 75: Normally the _SEC suffix would go at the end since it's most closely related to the value (20*60) and the concept of SEQUESTER_TIMEOUT_DEFAULT could (theoretically) be mapped to different units (e.g. _MIN or _MS).  The default for SEC would never be different than the default for MIN.  SEQUESTER_SEGMENT_SEC and SEQUESTER_SEGMENT_DEFAULT_SEC could be different values so the affinity is less (although I understand you're applying the _DEFAULT suffix to the parameter SEQUESTER_TIMEOUT).

I can't find anything that explicitly says this is the right way... but a quick Google CS search shows 29k hits for _DEFAULT_SEC and only ~300 for _SEC_DEFAULT, so... at least that's the prevailing convention.

controller/site_config/mac_addrs/9a02571e8f03/device_config.json
```diff
@@ -1,3 +1,8 @@
 {
-    "static_ip": "192.168.1.3"
+    "static_ip": "192.168.1.3",
+    "modules": {
+        "hold": {
+            "enabled": true
```
grafnu, line 5: e.g., even with "test_hold" flag not set, I would argue that this setting here should invoke the hold test just for this device.

henry54809: It does but I wanted to make it explicit so if the base config changes, hold will still be enabled.

forch/forchestrator.py
henry54809: I think it should be DEFAULT_SEQUESTER_TIMEOUT_SEC but that doesn't go along with SEQUESTER_SEGMENT_DEFAULT which was why I opted for SEQUESTER_TIMEOUT_SEC_DEFAULT to match. If you agree DEFAULT_SEQUESTER_TIMEOUT_SEC is better, I can change SEQUESTER_SEGMENT_DEFAULT to DEFAULT_SEQUESTER_SEGMENT as well.

controller/site_config/mac_addrs/9a02571e8f03/device_config.json
henry54809: It doesn't since the hold test is not part of the "configured tests". That's why I really wanted to remove the host_tests concept so only the module_config configures for what modules to run.

grafnu:
Yes, the DEFAULT_ prefix is better -- (just for my own curiosity) I did a
search in the Google codebase and DEFAULT_ prefix is more common than the
_DEFAULT suffix, although it's not quite conclusive (e.g., like 11M vs. 7M
hits).

henry54809:
PTAL.

grafnu (approved):

forch/device_report_server.py
```diff
@@ -12,9 +12,9 @@
 from forch.proto.shared_constants_pb2 import Empty, PortBehavior
 from forch.proto.devices_state_pb2 import DevicePortEvent
 
-ADDRESS_DEFAULT = '0.0.0.0'
-PORT_DEFAULT = 50051
-MAX_WORKERS_DEFAULT = 10
+DEFAULT_ADDRESS = '0.0.0.0'
```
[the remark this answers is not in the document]
henry54809: Normally I would but I have another PR for "incremental enable" that's based off of this already so I wanted to get things in quicker.

controller/site_config/mac_addrs/9a02571e8f03/device_config.json
henry54809: I verified with 1.10.5 and these configs that hold test doesn't run unless you specify the -k. Hold test isn't part of the configured tests so enabling it in config alone doesn't run the test.

PR #298, opened 2021-04-06
Adding incremental enable for FOT

henry54809:


proto/forch_configuration.proto
```diff
@@ -150,6 +150,9 @@ message OrchestrationConfig {
 
     // DUT will return to operational state if no DeviceReport is received within this timeout.
     int32 sequester_timeout_sec = 6;
+
+    // List of device macs that will be sequestered for testing
+    repeated string allowed_list = 7;
```
grafnu, line 155: "allowed" doesn't seem the right word here... I'd call this something like "target_macs" or "enabled_macs" or "sequestered_macs" -- the generalized concept would be to properties that could be applied to a list of device macs?

proto/forch_configuration.proto
```diff
@@ -150,6 +150,9 @@ message OrchestrationConfig {
 
     // DUT will return to operational state if no DeviceReport is received within this timeout.
     int32 sequester_timeout_sec = 6;
+
+    // List of device macs that will be sequestered for testing
```
grafnu, line 154: I'd represent this as a map not a list (even if the target object is empty).  Down the road, it's often very convenient to be able to add parameters/options to an entry, and that's trivial if it's already a map, but very messy if it's a list (and there's very little downside to doing it as a map initially).

forch/port_state_manager.py
```diff
@@ -207,7 +216,8 @@ def _handle_authenticated_device(self, mac, device_behavior, static):
             device_behaviors.setdefault(mac, DeviceBehavior()).CopyFrom(device_behavior)
 
             static_port_behavior = self._static_port_behaviors.get(mac)
-            if not self._sequester_segment or static_port_behavior == PortBehavior.cleared:
+            sequestering_allowed = _clean_mac(mac) in self._sequester_allowed_list
```
grafnu, line 219: One bit of functionality we need to control is the default behavior.  E.g., depending on configuration, a device that is listed in the set will either automatically be sequestered/tested or not.  The nominal case is that everything would be sequestered, this provision is to incrementally enable sequestering for controlled situations -- but we shouldn't get into a state where everything has to be listed always to be tested.

The easy thing to do here (IMHO) is in combination with the map concept below, just look up the "enabled" property in the map if the device is listed, and if it's not then go with a (configurable) default... or something like that.

henry54809: I had a bool flag for indicating whether to only test the list of devices but that's not pushed. Isn't that cleaner than a map? When other options are needed, those can go in a separate map under sequester config /  somewhere else that's uncoupled from the target macs.

grafnu:
I meant options per device... So they would be in a map indexed by device
mac.

henry54809:
I understood what you meant but I'm proposing to have a separate map for device options when we need, just like the following:
```
selective_sequestering: true
target_devices:
  - aa:bb:cc:dd:ee:01
  - aa:bb:cc:dd:ee:02
device_options:    <- When we need other device options
  aa:bb:cc:dd:ee:01:
    option1: value
  aa:bb:cc:dd:ee:03:
    option1: value
```
This is simpler given that you don't need to introduce enabled and default enabled concept in code; And it allows for the decoupling of config for devices that allow sequestering and config for other device options. If target devices and device options map were to be combined together, the resulting config can not be distinguish whether devices have specified enabled: false or left enabled blank. For example:
```
default_allow_sequestering: true
device_options:
  aa:bb:cc:dd:ee:01: 
    option1: value
    enable: true
  aa:bb:cc:dd:ee:02:
    option1: value
```
aa:bb:cc:dd:ee:02 will have enabled filled in as false by default (from protobuf) even though we actually want that device tested. This can't be solved by explicitly enabling device aa:bb:cc:dd:ee:02, then the default_allow_sequestering is not honored when that value is changed. This can be solved by introducing another enum for "enable", but that's one more concept to introduce for something that could be simpler.

grafnu:
Ok -- I had a deeper look, and I think the root of the problem is we
already have a device behavior mechanism that maps device mac to a set of
properties and does the inverse of this (you can mark a device as "cleared"
which will skip sequestering) -- so I'd posit that the explicit enabling
for sequestering should be the inverse of this existing behavior. Probably
need to check with yfd@ to make sure the logic follows correctly and plays
nice with "dynamic segment/role with static sequestering-enable". Which
seems to me would distill this down to:

   1. Add new configuration so that it can be configured to 'clear' by
   default rather than 'sequester' by default
   <https://github.com/grafnu/forch/blob/master/forch/port_state_manager.py#L210>
   (which is the current situation) -- or generally speaking "devices should
   start in XXX state by default"
   2. Then mark a device as 'pending' (or 'sequestered' might work since
   that state already exists) or something like that in the behavioral config
   file (rather than how it's currently marked 'cleared
   <https://github.com/grafnu/forch/blob/master/testing/python_lib/test_fot.py#L271>').
   Generally to mean ("except for this device which should start out YYY")

Basically, this mechanism allows the static configuration to specify the
"desired initial state" of a device (as 'cleared' or 'sequestered'), and
then if not specified, it falls back to the configured default (which would
in turn default to 'sequestered').  Logic should be fairly straight forward
and code change minimal since it's a simple deviation from the current
setup.

Part of the problem may be trying to conceptualize the values as booleans
("enabled") rather than an enum ('sequestered') -- and if the default value
of protobuf is confusing things then using enums instead should clear that
up right quick.  It's not really the case that "should this device be
sequestered" is a boolean, because it's really tri-state: { yes, no, and
follow-default }.

That aside, I would also be hard-pressed to buy the argument that two
independent lists of MAC addresses nearly side-by-side it better than one
-- or at the very least it's a big step back in terms of simplicity in my
opinion. Esp when they are fungible in that you could create the
"target_devices" list you have by filtering the map keys by the enabled
property, or generate the enabled property by checking for inclusion in the
list.  I'm less concerned about the complexity of the code since it would
be relatively minor and internalized, while the config file is externalized
and so generally more sticky and prone to accumulating debt.

forch/port_state_manager.py
didovesei: Is `allowed_list` functionally the same with `PortBehavior`?

We can explicitly set a device into the desired state (sequester, operational, etc.) by configuring the device's port behavior in the static behavior file.

https://github.com/faucetsdn/forch/blob/ae89a7776f687737f8e7c92e5dce8972768e18bb/proto/shared_constants.proto#L165

henry54809: This PR is obsolete now. There’s another one that’s merged.

didovesei:
> Ok -- I had a deeper look, and I think the root of the problem is we already have a device behavior mechanism that maps device mac to a set of properties and does the inverse of this (you can mark a device as "cleared" which will skip sequestering) -- so I'd posit that the explicit enabling for sequestering should be the inverse of this existing behavior. Probably need to check with yfd@ to make sure the logic follows correctly and plays nice with "dynamic segment/role with static sequestering-enable". Which seems to me would distill this down to:

Sorry I didn't see this comment when I left the previous one. It seems that the PortBehavior in static behavior file can achieve the same goal... We can discuss offline..

PR #299, opened 2021-04-07
Make Forch add operational VLANs to TAP port

didovesei:


testing/test_fot
```diff
@@ -17,7 +17,7 @@ YQ=venv/bin/yq
 
 echo DAQ spot checks
 
-docker exec forch-faux-4 ping -c10 192.168.1.0
+docker exec forch-faux-4 ping -c10 192.168.1.0 || true
```
grafnu, line 20: what's this for?  It's suppressing the test... ?

didovesei: Reverted the change for this line. I didn't notice that this device has a static IP and is not auto-sequstered. So this device will be set to operational state once Forch is started, which means ping is supposed to succeed at this point.

grafnu (approved):

PR #301, opened 2021-04-08
auto sequestering config

henry54809:


proto/shared_constants.proto
```diff
@@ -190,6 +190,12 @@ message PortBehavior {
     down = 0;
     up = 1;
   }
+
+  enum AutoSequestering {
+    default = 0;
+    true = 1;
```
grafnu, line 196: I think it'll be too confusing to use "true"/"false" as enum constants... so I'd posit these should be representative of the function... so something like "ENABLED" or "SKIP" respectively -- whatever.  But, I think in this context boolean == bad.

proto/forch_configuration.proto
```diff
@@ -150,6 +150,9 @@ message OrchestrationConfig {
 
     // DUT will return to operational state if no DeviceReport is received within this timeout.
     int32 sequester_timeout_sec = 6;
+
+    // Default value for DeviceBehavior.auto_sequestering
+    bool default_auto_sequestering = 7;
```
grafnu, line 155: Should be of the same type as the behavior, I think... so PortBehavior.AutoSequestering so it's very clear how it matches.

forch/port_state_manager.py
```diff
@@ -155,6 +158,13 @@ def handle_device_placement(self, mac, device_placement, static=False):
 
         return self._handle_disconnected_device(device_placement)
 
+    def _resolve_auto_sequester_config(self, auto_sequestering):
+        if auto_sequestering == PortBehavior.AutoSequestering.true:
```
grafnu, line 162: with the appropriate type changes this could be simplified to:

auto_sequestering = self._default_auto_sequestering if auto_sequestering == unknown

or something like that.

proto/forch_configuration.proto
henry54809: But that means you can have a “default” as the value of default. What’s the default value for the “default” of default?

proto/shared_constants.proto
henry54809: I can change these to enabled and disabled but I think the default_auto_sequester should still be a bool. See reasoning below...

grafnu:
It would be functionality neutral with the field not existing at all.  We
get to decide.  I don't think there's a "wrong" answer, as such -- but it
could be...

Same as false  (so field not present is the same)
Same as current behavior (which is to automatically sequester)
Same as intended behavior (which is to automatically sequester)
Same as "safe" behavior (which is not to automatically sequester)

So, I would lean towards default would be not to sequester -- so equivalent
to boolean false.

henry54809:
Updated. PTAL

grafnu (approved):

PR #305, opened 2021-04-14
configurable infracted state

henry54809:


didovesei (approved):

proto/shared_constants.proto
```diff
@@ -93,6 +93,10 @@ message DVAState {
 
     // infracted
     infracted = 5;
+
+    // operational
+    // TODO.
```
didovesei, line 98: NIT: Might need to be removed?

proto/shared_constants.proto
```diff
@@ -159,6 +163,26 @@ message PortType {
...
+/*
+ * TestResult
+ * TODO: Replace this enum with SessionResult in DAQ.
+ */
+
+/*
+ * Session result codes.
+ */
+message SessionResult {
+  SessionResult.ResultCode code = 1;
```
didovesei, line 175: If the message contains a data field, then I don't think it would be a good idea to define it in shared_constants.proto...

forch/port_state_manager.py
```diff
@@ -22,63 +21,93 @@ def register(func):
...
+            PortBehavior.Behavior: {
+                PortBehavior.Behavior.cleared: OPERATIONAL,
+                PortBehavior.Behavior.deauthenticated: UNAUTHENTICATED
+            }
         },
     }
 
+    class StateCallbacks:
+        """Wrapper for state machine callbacks"""
+        unauthenticated_state = None
```
grafnu, line 58: seems odd to have a whole class for this rather than just a map indexed by the relevant constants.... not a big deal but this seems unnecessarily verbose.

proto/forch_configuration.proto
```diff
@@ -157,6 +157,14 @@ message OrchestrationConfig {
...
     // Default value for DeviceBehavior.auto_sequestering
     PortBehavior.AutoSequestering default_auto_sequestering = 8;
+
+    message TestResultDeviceStateTransition {
+      SessionResult.ResultCode result = 1;
+      DVAState.State device_state = 2;
+    }
+
+    // Specify resulting device states from test results
+    repeated TestResultDeviceStateTransition test_result_device_state = 9;
```
grafnu, line 167: test_result_device_states  (plural), since it's a collection

proto/shared_constants.proto
grafnu: is the field actually used?  Seems like for this just need the constants?

topo/fot/forch/forch.yaml
```diff
@@ -20,6 +20,11 @@ orchestration:
     port_description: TAP
     sequester_timeout_sec: 360
     default_auto_sequestering: disabled
+    test_result_device_state:
```
grafnu, line 23: hard to tell from the code, but are there reasonable (functional) defaults for these? What happens if it's not defined or there's an unconfigured result?

topo/fot/forch/forch.yaml
```diff
@@ -20,6 +20,11 @@ orchestration:
...
+      - result: FAILED
+        device_state: infracted
+      - result: PASSED
```
grafnu, line 26: I'd say either { test_result, device_state } or { result, state } for consistency.

proto/shared_constants.proto
henry54809: I copied this directly from your PR. I didn't know what it was intended so I left it like that.

topo/fot/forch/forch.yaml
henry54809: Defined here is what the default is in code. The default is the state machine dict on PortStateMachine, and anything defined here overwrites the state machine transitions.

proto/shared_constants.proto
henry54809: I left the TODO in there so we can remove static_operational and dynamic_operational in the future

henry54809:
Updated. PTAL

topo/fot/forch/forch.yaml
grafnu: so what's the purpose of this bit then?  Is it actually testing anything?  I'm not sure how to get away with a simple integration test change for this...   or was this just meant as an example config to point people at?

henry54809: There’s another test that tests the opposite behavior in unit test.

grafnu:
But what's the purpose of these configs, if they are the default?

topo/fot/forch/forch.yaml
henry54809: It is to test config ingestion, state machine state transition overwrites. The config still overwrites what is in the default even though they’re the same. By specifying the default and the result doesn’t change shows the introduction of the config doesn’t break anything. The unit test tests the other side; by specifying the opposite of default, the resulting device states changes.

grafnu:
But it doesn't, because the load config could do absolutely nothing and the
result would be the same... Also that seems like it would be better for a
unit test (because it really is isolated to just that function).

This would be most useful if it showed how to partially override the config
to change behavior, but I don't see how to easily do that without breaking
everything, and I don't think it's worth it.

Anyway, I'd take it out rather than leave it in it's current form because I
think it's more confusing as not to have it there.

topo/fot/forch/forch.yaml
henry54809: By this test alone, the config could be doing nothing and still not change results. With the unit tests and this test, it shows the config logic works as intended whether the config specifies nothing, or specifies devices go into infracted or specifies devices go into operational. Taking it out removes the testing of 1 of these scenarios.

grafnu:
Help me understand here... What exactly breaks (like where would the error
whole up) if this bit were omitted?

grafnu (approved):

forch/port_state_manager.py
```diff
@@ -125,9 +155,21 @@ def __init__(self, device_state_manager=None, varz_updater=None,
         self._sequester_timer = {}
         self._lock = threading.RLock()
         self._logger = get_logger('portmgr')
+        self._state_callbacks = self._build_state_callbacks()
+        self._state_overwrites = {}
         if sequester_config:
             self._sequester_segment = sequester_config.sequester_segment
             self._sequester_timeout = sequester_config.sequester_timeout_sec
+            if sequester_config.test_result_device_states:
+                dict_maps = [(entry.result, entry.device_state)
```
grafnu, line 164: entry.test_result?

PR #308, opened 2021-04-23
Move DVA state calculation to PortStateManager

didovesei:


forch/port_state_manager.py
```diff
@@ -395,3 +396,28 @@ def _update_device_state_varz(self, mac, device_state):
...
+
+    def _get_dva_state(self, switch, port):
+        mac = self._placement_to_mac.get((switch, port))
+        if not mac:
+            if self._orch_config and self._orch_config.unauthenticated_vlan:
+                return DVAState.unauthenticated
+            return DVAState.initial
+
+        state_machine = self._state_machines.get(mac)
+        if not mac:
```
henry54809, line 413: if not state_machine?

forch/port_state_manager.py
```diff
@@ -395,3 +396,28 @@ def _update_device_state_varz(self, mac, device_state):
     def _update_static_vlan_varz(self, mac, vlan):
         if self._varz_updater:
             self._varz_updater.update_static_vlan_varz(mac, vlan)
+
+    def get_dva_state(self, switch, port):
```
henry54809, line 400: also why can't the fcoll get the device state directly from varz? It seems to be setting most of the states twice and duplicating the logic.

forch/port_state_manager.py
didovesei: Fixed

forch/port_state_manager.py
didovesei: So the varz is only set for learned devices while `get_dva_state` should return the dva state for all the access ports including those that have no devices connected. That's why they are processed in different places.

Varz is generated by the prometheus client and there may be delays for the varz to show up after they are set. As PortStateManager is the authorative source of the port states, fcoll should get the dva states directly from the manager instead of the varzs.

PR #311, opened 2021-05-03
Handle timeouts for DTS gRPC and add in testing w/ vxlan.

grafnu:


henry54809 (approved):

forch/device_report_client.py
```diff
@@ -56,27 +59,42 @@ def _connect(self, mac, vlan, assigned):
...
+
+    def disconnect(self, mac):
+        with self._lock:
+            session = self._mac_sessions.get(mac)
+            if session:
+                session.cancel()
+                self._mac_sessions.pop(mac)
+                self._logger.info('Device %s disconnected', mac)
+            else:
+                self._logger.warning('Disconnecting unconnected device %s', mac)
```
anurag6, line 75: NIT: Should this log message be more on the lines of 'Can't disconnect unconnected device' or 'Device not connected' since the method doesn't actually do anything if it is disconnected?

anurag6 (approved):

forch/device_report_client.py
grafnu: Yup -- updated.

PR #317, opened 2021-06-02
FOT cleanup after terminal state, and port debounce improvements

grafnu:


forch/faucet_event_client.py
```diff
@@ -151,22 +151,33 @@ def _debounce_port_event(self, event, port, active):
...
-                                lambda: self._handle_debounce(event, port, active))
-        timer.start()
-        self._port_timers[state_key] = timer
+        with self._buffer_lock:
+            if state_key in self._port_timers:
+                self._logger.debug('Port cancel %s', state_key)
+                self._port_timers[state_key].cancel()
+                if active:
+                    self._logger.info('Ignoring spurious port down event')
+                    self._port_timers.pop(state_key)
```
henry54809, line 160: This line is extra since the timer is also popped in handle_debounce. and shouldn't the line above say "port up" event?

henry54809: Or is it missing a return?

henry54809 (approved):

PR #318, opened 2021-06-03
Handle exceptions and reduce severity of logs

anurag6:
Reduces severity of attempt logs to prevent spam in main log file. 

Handles exceptions when attempting to retrieve metrics.

### DIFF
--- a/forch/forchestrator.py
+++ b/forch/forchestrator.py
@@ -484,7 +484,7 @@ class Forchestrator(VarzUpdater, OrchestrationManager):
 
     def _get_varz_config(self):
         metrics = self._varz_collector.retry_get_metrics(
-                self._faucet_prom_endpoint, _TARGET_FAUCET_METRICS)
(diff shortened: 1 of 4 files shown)


forch/forchestrator.py
```diff
@@ -484,7 +484,7 @@ def _register_handlers(self):
 
     def _get_varz_config(self):
         metrics = self._varz_collector.retry_get_metrics(
-            self._faucet_prom_endpoint, _TARGET_FAUCET_METRICS)
+                self._faucet_prom_endpoint, _TARGET_FAUCET_METRICS)
```
grafnu, line 487: is this right?  I thought the hanging indent was supposed to be 4... ?

grafnu (approved):

forch/forchestrator.py
anurag6: Yeah, that's an oversight. Correcting.

PR #319, opened 2021-06-03
Retry mechanism for testing service connection

grafnu:


forch/device_report_client.py
```diff
@@ -90,13 +90,19 @@ def _convert_and_handle(self, mac, progress):
...
         try:
             for progress in session:
+                should_retry = False
                 if self._convert_and_handle(mac, progress):
                     self._logger.warning('Terminal state reached for %s', mac)
             self._logger.error('Progress complete for %s', mac)
         except Exception as e:
             self._logger.error('Progress exception: %s', e)
+        self.disconnect(mac)
+        if should_retry:
```
henry54809, line 103: It looks like the session progress can be empty, thus session needs to be restarted. Shouldn't we fix the root cause of why session can be empty coming from DAQ?

grafnu:
The root cause was that the client tried to connect to a not-yet-started
server, so it wasn't that the session was empty, it was that it threw an
exception.  The actual path through the code to a retry was through the
exception handler.  I've never seen the case where session was empty, and
I'm not even sure how that would happen.

henry54809:
What about just let forch wait until the channel is ready? https://grpc.github.io/grpc/python/grpc.html#channel-ready-future

grafnu:
Updated to use channel ready future, PTAL.

henry54809 (approved):

PR #321, opened 2021-06-07
Working with latest DTS setup

grafnu:


forch/device_report_client.py
```diff
@@ -23,20 +25,16 @@
     pass
 
 
-DEFAULT_SERVER_ADDRESS = '127.0.0.1'
-DEFAULT_SERVER_PORT = 50051
 CONNECT_TIMEOUT_SEC = 60
+LOCAL_ADDRESS = '127.0.0.1'
```
henry54809, line 29: This is not needed now right?

forch/endpoint_handler.py
```diff
@@ -0,0 +1,93 @@
...
+        server_port = DEFAULT_SERVER_PORT
+        address = f'{target_ip}:{server_port}'
+        self._logger.info('Proxy requests to %s', address)
+        channel = grpc.insecure_channel(address)
+        self._stub = server_grpc.SessionServerStub(channel)
+
+    def process_endpoint(self, endpoint):
+        """Handle an endpoint request"""
+        self._logger.info('Process request for %s', endpoint.ip)
+        self._stub.ConfigureEndpoint(endpoint)
```
henry54809, line 41: I am confused about this setup. Is this for setuping up a vxlan tunnel to DAQ? then why is there the EndpointServer here to run a command... 
Also ConfigureEndpoint was previously defined in DAQ's grpc; it has since been removed so there is some disconnect somewhere.

forch/device_report_client.py
grafnu: removed

forch/endpoint_handler.py
grafnu: Right, I remember now. This is needed for some for some of the testing setups. I added it back into DAQ and a (lengthy) comment at the top of this file explaining what it's used for.

PR #324, opened 2021-06-29
Use a different process for each device's grpc connection

henry54809:
GRPC only maintains 1 single channel (connection to the k8s IBL) in each process so LB across all testing devices can't happen. This PR, with no functional change, ensures each device can have its own connection to the IBL for proper LB.

### DIFF
--- a/forch/device_report_client.py
+++ b/forch/device_report_client.py
@@ -1,10 +1,12 @@
 """Server to handle incoming session requests"""
 
+import signal
 import threading
+import multiprocessing as mp
+
 import grpc
 
 import forch.endpoint_handler as endpoint_handler
-
 from forch.proto.shared_constants_pb2 import PortBehavior
 from forch.proto.devices_state_pb2 import DevicesState
 from forch.base_classes import DeviceStateReporter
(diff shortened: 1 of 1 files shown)


forch/device_report_client.py
```diff
@@ -99,18 +122,21 @@ def _convert_and_handle(self, mac, progress):
...
-            self._logger.error('Progress exception: %s', e)
-        self.disconnect(mac)
+    def _process_progress_q(self):
+        while True:
+            mac, progress = self._progress_q.get(block=True)
+            if not mac:  # device client shutdown
+                break
+            try:
+                if not progress or self._convert_and_handle(mac, progress):
+                    print('Progress complete for %s' % mac)
```
anurag6, line 132: Debug prints left by mistake?

anurag6 (approved):

forch/device_report_client.py
```diff
@@ -50,30 +55,48 @@ def __init__(self, result_handler, target, unauth_vlan, tunnel_ip):
 
     def start(self):
         """Start the client handler"""
-        grpc.channel_ready_future(self._channel).result(timeout=CONNECT_TIMEOUT_SEC)
-        self._stub = SessionServerStub(self._channel)
+        # Context may be set already
```
grafnu, line 58: I'm not sure what to make of this comment. Can you expand it a bit or position it better?  Is this to say that if the context is already set then this will throw a RuntimeError?  If so, I'd either clarify that in the comment or move to the except clause.

grafnu (approved):

PR #329, opened 2021-07-13
Feature/custom faux mac prefix

henry54809:


.github/workflows/main.yaml
```diff
@@ -73,8 +73,12 @@ jobs:
       - name: run python tests
         run: testing/python_test test_fot
       - name: setup fot stack
+        env:
+          FAUX_MAC_PREFIX: 9a:99:57:1e
```
grafnu, line 77: is it possible to make this an argument to setup_stack?  Using an env variable is a different point-of-configuration than what's already employed (command line args).  Then maybe setup_stack could set the env variable internally if it needs to cascade down to other things?

bin/run_test_set
```diff
@@ -10,15 +10,18 @@ if [ -z "$TEST_SET" ]; then
...
 OUT_DIR=test_out/
 rm -rf $OUT_DIR
 mkdir -p $OUT_DIR
 echo Writing output to $OUT_DIR
 
-GOLDEN_RESULTS=etc/test_$TEST_SET.out
+GOLDEN_RESULTS=$OUT_DIR/golden_test_$TEST_SET.out
 TEST_RESULTS=$OUT_DIR/test_$TEST_SET.out
-
+sed "s/9a:02:57:1e/$FAUX_MAC_PREFIX/g" etc/test_$TEST_SET.out > $GOLDEN_RESULTS
```
grafnu, line 23: what's the need for this? Usually we like to keep tests unparameterized as much as possible to try and cleanly separate out code from test.

etc/start_faux
```diff
@@ -52,8 +52,8 @@ ip route
 if [ -n "${options[dnsmasq]}" ]; then
     echo Configuring dnsmasq...
     echo dhcp-range=192.168.1.10,192.168.1.255 >> /etc/dnsmasq.conf
-    echo dhcp-host=9a:02:57:1e:8f:01,192.168.1.1 >> /etc/dnsmasq.conf
-    echo dhcp-host=9a:02:57:1e:8f:02,192.168.1.2 >> /etc/dnsmasq.conf
+    echo dhcp-host=$FAUX_MAC_PREFIX:8f:01,192.168.1.1 >> /etc/dnsmasq.conf
```
grafnu, line 55: Is it possible to use a command-line arge to start_faux?  There's already the mechanism for command-line options w/ args...

testing/test_fot
```diff
@@ -76,30 +76,30 @@ vlan=$($YQ '.dps."nz-kiwi-t2sw1".interfaces."1".native_vlan' $BEHAVIORAL_CONFIG)
 echo $(( (vlan > 271) + (vlan > 276) )) | tee -a $TEST_RESULTS
 echo $(( (vlan > 271) + (vlan > 276) )) | tee -a $TEST_RESULTS
 
-monitor_daq_log "Send device result 9a:02:57:1e:8f:01 passed"
-monitor_daq_log "Send device result 9a:02:57:1e:8f:02 failed"
-monitor_daq_log "Target device 9a02571e8f03 test hold running"
-monitor_daq_log "Target device 9a02571e8f05 test hold running"
+monitor_daq_log "Send device result $FAUX_MAC_PREFIX:8f:01 passed"
```
grafnu, line 79: similar to other commend, generally like to avoid parameterizing tests unless absolutely necessary. Can all the FOT tests just be run with a fixed alternate prefix, avoiding the need to parameterize here?

.github/workflows/main.yaml
henry54809: bin/setup_stack has too many positional arguments already so I opted for an env variable.

grafnu:
I think you're confusing positional arguments with flag arguments that just
happen to be in a specific order (but isn't strictly positional). I don't
think opting for the env var is the right answer here, I'd either 1) just
pile on to the existing how-it's-broken and so it's not bifurcating the
problem, or 2) we take the little bit of time to clean up the current
mechanism so it's not so persnickety about argument ordering.  Personally
I'd just go with option #1, but if you feel strongly about it and/or don't
want to deal with it (valid position), then I can take a crack at fixing it
up before finalizing this PR (should be a fairly easy thing I can crank
through today).

henry54809:
PTAL

bin/run_controller
```diff
@@ -153,6 +153,10 @@ if [[ -z $vxlan ]]; then
     docker exec $CONTAINER ip link set data0 up
 fi
 
+if [ $CLEAN_MAC_PREFIX != "$DEFAULT_CLEAN_MAC_PREFIX" ]; then
+    docker exec $CONTAINER /root/bin/rename_site_config $CLEAN_MAC_PREFIX
```
grafnu, line 157: I think this should go inside controller_go -- basically just pass the CLEAN_MAC_PREFIX in (as per the docker_envs use above), and then inside of the controller sort out what to do about it. Generally better to avoid the "docker exec" solution because it's a lot harder to debug/diagnose/trace -- the only reason it's done for the data0 interface is because it has to be done externally after the container is created (effective race condition).

bin/run_faux
```diff
@@ -2,13 +2,16 @@
...
 docker_args=
 container_prefix=forch-faux
 image=forch/faux
 net_opt=--net=none
 port_num=0
 
+source bin/stack_functions
+set_faux_mac_prefix
+
+intf_mac=$FAUX_MAC_PREFIX:8f:00
```
grafnu, line 14: what happens if FAUX_MAC_PREFIX is not defined?  Doesn't this need a default somewhere?

henry54809: set_faux_mac_prefix defines the default

grafnu (approved):

PR #331, opened 2021-07-15
Manual fot config changes

henry54809:
Planning on using a third party library (dateutil) to do the date parsing since the built in library doesn't support the full ISO8601 standard.

### DIFF
--- a/proto/devices_state.proto
+++ b/proto/devices_state.proto
@@ -31,6 +31,8 @@ message DeviceBehavior {
     PortBehavior.Behavior port_behavior = 3;
     PortBehavior.AutoSequestering auto_sequestering = 4;
   }
+  // ISO8601 timestamp string for manual sequestering
+  string manual_sequestering_timestamp = 6;
 }
 
 message SegmentsToVlans {
--- a/topo/fot/forch/behaviors.yaml
+++ b/topo/fot/forch/behaviors.yaml
@@ -8,6 +8,7 @@ device_mac_behaviors:
   "9a:02:57:1e:8f:03":
     segment: INFRA
     auto_sequestering: enabled
+    manual_sequestering_timestamp: "2021-07-15T18:31:12"
   "9a:02:57:1e:8f:04":
     segment: INFRA
   "9a:02:57:1e:8f:05":
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -16,3 +16,4 @@ setuptools
 watchdog
 wheel
 yq
+python-dateutil==2.8.2


proto/devices_state.proto
```diff
@@ -31,6 +31,8 @@ message DeviceBehavior {
     PortBehavior.Behavior port_behavior = 3;
     PortBehavior.AutoSequestering auto_sequestering = 4;
   }
+  // ISO8601 timestamp string for manual sequestering
+  string manual_sequestering_timestamp = 6;
```
grafnu, line 35: this isn't necessarily going to be "manual" so I'd leave that qualifier out. Mechanically, what does this mean? Maybe next_sequestering_timestamp?

topo/fot/forch/behaviors.yaml
```diff
@@ -8,6 +8,7 @@ device_mac_behaviors:
   "9a:02:57:1e:8f:03":
     segment: INFRA
     auto_sequestering: enabled
```
grafnu, line 10: somewhat begs the question around what does "auto_sequestering" mean -- because there's all sorts of "auto" -- not something we need to change/address now, more just pontificating.

grafnu:
Do you know if that package is in rapture?  That'd be my only concern....
using it is fine, but if it's not already there then somebody would need to
import it, yada yada...

proto/devices_state.proto
henry54809: I want to differentiate the auto and manual. This sequestering timestamp may not be the actual next sequestering timestamp since between the time the config is read and the timestamp, device can go offline and online again which will trigger the auto sequestering. I think scoping this timestamp to manual is a better option.

topo/fot/forch/behaviors.yaml
henry54809: the package is readily available.

grafnu:
Differentiating is fine/good, but it's not necessarily manual. What happens
if that field is automatically generated?  You're only thinking about the
one specific use case of "manual" -- but that doesn't describe what it
does, it describes how you think it'll be used. Generally speaking, fields
should be named for what they do, since that's immutable unlike the use of
it.  auto_sequestering as a field is a bit better, because it can be
interpreted as the behavior of how the system will automatically do
something (e.g., never, periodically, event_up, etc...). "manual" doesn't
work the same since computer systems don't *manually* do things. Some other
alternatives: directed_, sporadic_, triggered_, scheduled_.

henry54809:
@grafnu  I like scheduled_sequestering; I'll replace the field with that.

PR #332, opened 2021-07-20
Feature/manual fot

henry54809:


forch/device_report_client.py
```diff
@@ -142,3 +142,9 @@ def process_port_assign(self, mac, vlan):
         with self._lock:
             self._mac_assigned_vlan_map[mac] = vlan
             self._process_session_ready(mac)
+
```
grafnu, line 145: How is this conceptually different than the previous function?   When would they be used differently?  If they're more the same than not, I would advocate for changing the name of the existing function to be more appropriate to the generalized use rather than creating a duplicate function "just because it's used from a different place"...

forch/port_state_manager.py
```diff
@@ -188,6 +192,28 @@ def handle_static_device_behavior(self, mac, device_behavior):
             self._auto_sequester[mac_lower] = auto_sequester
             if device_behavior.segment:
                 self.handle_device_behavior(mac_lower, device_behavior, static=True)
+            sequester_scheduled = self._scheduled_sequester_timer.get(mac_lower)
+            if device_behavior.scheduled_sequestering_timestamp and not sequester_scheduled:
+                try:
```
grafnu, line 197: extract function here -- it would be nice to keep all the tz/date stuff isolated in one place, and keep the handle_static_device_bahavior function shorter.

forch/port_state_manager.py
```diff
@@ -188,6 +192,28 @@ def handle_static_device_behavior(self, mac, device_behavior):
...
+                                       device_behavior.scheduled_sequestering_timestamp)
+                    return
+                local_tz = dateutil.tz.tzlocal()
+                if not parsed.tzinfo:
+                    parsed = parsed.replace(tzinfo=local_tz)
+                if parsed >= datetime.now(local_tz):
+                    time_diff = parsed - datetime.now(local_tz)
+
+                    def handler():
+                        self._handle_scheduled_sequstering(mac_lower)
```
grafnu, line 210: did you think through the multithreaded processing part here, to make sure the appropriate locks are in place when multiple threads are doing whatever?

forch/port_state_manager.py
```diff
@@ -188,6 +192,28 @@ def handle_static_device_behavior(self, mac, device_behavior):
...
+            sequester_scheduled = self._scheduled_sequester_timer.get(mac_lower)
+            if device_behavior.scheduled_sequestering_timestamp and not sequester_scheduled:
+                try:
+                    parsed = dateutil.parser.parse(device_behavior.scheduled_sequestering_timestamp)
+                except dateutil.parser.ParserError:
+                    self._logger.error("Failed to parse scheduled sequestering timestamp: %s.",
+                                       device_behavior.scheduled_sequestering_timestamp)
+                    return
+                local_tz = dateutil.tz.tzlocal()
+                if not parsed.tzinfo:
```
grafnu, line 204: NB: There is a race (?) condition here where the timeout won't happen if the system happens to be restarting when the sequestered is scheduled.  but -- I don't care about that so I don't think it's worth addressing -- just throwing the comment in for completeness. Maybe add a TODO or something acknowledging this.

forch/port_state_manager.py
```diff
@@ -340,6 +366,13 @@ def _handle_session_result(self, mac, session_result):
     def _handle_unauthenticated_state(self, mac):
         self._update_device_state_varz(mac, DVAState.unauthenticated)
 
+    def _handle_scheduled_sequstering(self, mac):
```
grafnu, line 369: I think there might be an issue here where multiple timers can get set for the same mac, and then the first one will fire even if the date was changed to later?   Likely doesn't matter, but maybe just add a check here to only actually do the sequester stuff once.  Or at least something to consider.  At the very least add a comment or note saying that this is a known "problem" in the code.

forch/port_state_manager.py
henry54809: The multiprocess is contained in the device report client so it doesn't affect anything here; the creation of the timer is done with a lock so each device should only have 1 timer active at any time. It does look like I ignore any changes to schedule time change, so I'll make a change to update this.

henry54809:
PTAL

grafnu (approved):
Other than Stickler LGTM

PR #334, opened 2021-07-26
Remove scheduled sequesters if timestamps in config are removed

henry54809:


forch/port_state_manager.py
```diff
@@ -193,9 +193,9 @@ def handle_static_device_behavior(self, mac, device_behavior):
             if device_behavior.segment:
                 self.handle_device_behavior(mac_lower, device_behavior, static=True)
             scheduled_sequester = self._scheduled_sequester_timer.get(mac_lower)
+            if scheduled_sequester:
```
grafnu, line 196: I think this will still leave the (cancelled) timer object in the map which makes the check for "is timer in map" not meaningful. Maybe replace the "get" with a pop w/ default (or was it remove w/ default?) to also pull it out of the map.

grafnu (approved):

PR #337, opened 2021-08-20
multi vxlan

henry54809:
Need to make 1 more change on DAQ and merge DAQ in first.


forch/faucet_event_client.py
```diff
@@ -110,7 +110,7 @@ def _valid_event_order(self, event):
         assert self._last_event_id, '_last_event_id undefined, check for initialization errors'
         event_id = int(event['event_id'])
         if event_id <= self._last_event_id:
-            self._logger.debug('Outdated faucet event #%d', event_id)
+            self._logger.info('Outdated faucet event #%d', event_id)
```
henry54809, line 113: Will revert this.

proto/endpoint_server.proto
```diff
@@ -0,0 +1,29 @@
...
+    // IP address of tunnel endpoint.
+    string ip = 1;
+  
+    // UDP port for tunnel.
+    int32 port = 2;
+  
+    // Tunnel key index.
+    int32 vni = 3;
+
+    // Port on t1 switch
```
grafnu, line 18: nit: don't reference t1 here (it's a very specific implementation) -- maybe "egress"?

proto/endpoint_server.proto
```diff
@@ -0,0 +1,29 @@
...
+    int32 tap_port = 4;
+  }
+
+/*
+ * Device testing session server endpoint.
+ */
+ service EndpointServer {
+     // Configure Vxlan interface
```
grafnu, line 26: nit: indent

also, not sure the comment is helpful/accurate -- at this level (where the comment is), the interface may very well not be vxlan.  Personally I'd just remove it -- excess comments often seem like a good idea at the time, but in the long run they end up being outdated.

proto/endpoint_server.proto
```diff
@@ -0,0 +1,29 @@
...
+    rpc ConfigureInterface(Endpoint) returns (Empty) {}
+    rpc CleanupInterface(Endpoint) returns (Empty) {}
```
grafnu, line 28: a pontificate that I wouldn't bother addressing but I need to ruminate... Ideally I think this would be a streaming RPC such that if the client were to die for some reason, there would be an implicit "cleanup" on termination of the connection. Not worth fixing since this isn't code I think we should invest a lot in at this point, but there's likely "garbage collection" issues introduced by this!

forch/endpoint_handler.py
```diff
@@ -13,54 +13,100 @@
...
 
 DEFAULT_SERVER_PORT = 50111
 DEFAULT_BIND_ADDRESS = '0.0.0.0'
 DEFAULT_VXLAN_PORT = 4789
-DEFAULT_VXLAN_VNI = 0
-
+BASE_T1SW_PORT = 29
 VXLAN_CMD_FMT = 'ip link add %s type vxlan id %s remote %s dstport %s srcport %s %s nolearning'
-
+T1_SW1 = 'nz-kiwi-t1sw1'
```
grafnu, line 30: Add TODO that this should be configurable.  It's only a constant used by testing, but that's all this code is currently used for -- so I don't think it's worth actually making it configurable now, but at least acknowledge with TODO.

Or, at least investigate if it's reasonably trivial to extract this from the structural_config_file below.  I know it's a bit hacky, but I think a heuristic like "identify switch with TAP port description -- choose the one with lowest dp_id" might work well.

Up to you -- don't think it'll actually matter in the long run so do whatever.

forch/endpoint_handler.py
```diff
@@ -13,54 +13,100 @@
...
 
 
 class EndpointHandler:
     """Class to handle endpoint updates"""
 
-    def __init__(self, target_ip):
+    def __init__(self, target_ip, structural_config_file):
         self._logger = get_logger('endpproxy')
+        self._lock = threading.RLock()
+        self._mac_t1sw1_port = {}
```
grafnu, line 40: Don't encode "t1sw1" here (variable name) -- at the very least, it's requiring somebody who's trying to understand this code to then understand the switch hierarchy. I'd just replace all instances of t1sw1 with "egress" or something more semantically meaningful.

forch/endpoint_handler.py
```diff
@@ -13,54 +13,100 @@
...
+                t1sw_port = max(self._freed_t1sw_ports)
+                self._freed_t1sw_ports.remove(t1sw_port)
+            else:
+                t1sw_port = self._next_t1sw_port
+                self._next_t1sw_port += 1
+            self._mac_t1sw1_port[mac] = t1sw_port
+            with open(self._structural_config_file, 'r') as file:
+                structural_config = yaml.safe_load(file)
+                structural_config['dps'][T1_SW1]['interfaces'][t1sw_port] = {
+                    'description': 'TAP',
```
grafnu, line 65: nit: _ideally_ (and not necessary to change), this info (description and tagged_vlans) should be copied from the already present TAP port that's there. But, that might be unnecessary work given the limited lifespan of this code.

At the very least please extract to function (the port dict info), so that it's conceptually separated and then add a TODO at the function level of how the information is determined.

forch/endpoint_handler.py
```diff
@@ -13,54 +13,100 @@
...
-
-class EndpointServicer(SessionServerServicer):
+    def free_endpoint(self, mac: str):
+        with self._lock:
+            self._logger.info('Process request to free endpoint for %s', mac)
+            freed_port = self._mac_t1sw1_port.pop(mac, None)
+            if not freed_port:
+                return
+            self._freed_t1sw_ports.add(freed_port)
+            with open(self._structural_config_file, 'r') as file:
```
grafnu, line 92: Can you extract the parts that deal with the structural config to a separate function?  It's mixing concepts here (endpoint management, with manipulating the file) so I think it'd be cleaner to factor out.

Personally I'd extract to one function (combining with above) that accepted a lambda for "do a think to this interface" that in one case would be setting it to something, and in the other case setting it to empty.  All the helper stuff around the file (safe_load, dump) would ideally only be in one place.

grafnu (commented):
A few pontifications about what "ideal" would be, but not required to change.

And then a few light refactoring things to make it a bit more clean.

proto/endpoint_server.proto
henry54809: The stream is already on the session server so we know when a connection is terminated. During the implicit cleanup of these sessions, the interfaces used are explicitly cleaned up. I don't think we need to maintain another stream from session server to the endpoint handler

forch/endpoint_handler.py
henry54809: There is already port 28 named egress that goes to corp. Is there another named that can be used for these ports?

grafnu:
Well, I think we're looking for a name of the switch (t1sw1) not the port,
right?  That's why I was thinking of "egress" but you're right that it's
confusing. How about "pri" (as in primary)?  That's the term owca used,
which I think is descriptive. There's basically "primary" and "access"
switches. Although I think that might be breaking the abstraction (b/c its
getting into the internals of the switching fabric). At one level, we're
also just talking about a "switch" port, so we could be doing a t1sw1 -->
switch. My concern there would be that "switch" is too generic. Maybe "tap"
works well?  so _mac_tap_port = {}? Or stap if you wanted to have a
somewhat more unique term (switch test access port, or similar).

PR #349, opened 2021-11-11
Fix broken tests

grafnu:


testing/python_lib/build_config.py
```diff
@@ -246,7 +265,8 @@ def main(argv):
     else:
         raise Exception('Unkown topology type: %s' % topo_type)
 
-    config_map = proto_dict(faucet_config)
+    config_map = cleanup_config(proto_dict(faucet_config))
```
henry54809, line 268: Can the fix be in proto_dict instead? Or for some situations, it’s necessary to have str keys?

testing/python_lib/build_config.py
```diff
@@ -201,6 +202,24 @@ def create_corp_faucet_config(self):
         return FaucetConfig(dps=dps, version=2)
 
 
+def cleanup_keys(map_in):
+    """proto_dict converstaion converts int keys to strings, which causes problems with faucet."""
```
henry54809, line 206: Is “converstaion” conversion?

testing/python_lib/build_config.py
grafnu: There are definitely cases where str keys are valid, so I can't think of any easy way to make it more general without being overly complicated (as in, dynamically investigating the proto to see what it's supposed to be and then patching it up).

testing/python_lib/build_config.py
grafnu: updated

testing/python_lib/build_config.py
```diff
@@ -201,6 +202,24 @@ def create_corp_faucet_config(self):
...
+    return map_out
+
+
+def cleanup_config(config_map):
+    """cleanup the necessary dict entries"""
+    for dp_id in config_map['dps']:
+        bad_map = config_map['dps'][dp_id]['interfaces']
+        config_map['dps'][dp_id]['interfaces'] = cleanup_keys(bad_map)
+    if 'vlans' in config_map:
+        config_map['vlans'] = cleanup_keys(config_map['vlans'])
```
henry54809, line 219: It's really not necessary to have another function for this, you can replace cleanup_keys with this line `dict(iter((int(k), v) for k, v in config_map['vlans'].items()))`

henry54809 (approved):

grafnu:
I think using the dict/iter approach is good (better than explicit for
loop), but I think the function itself is still worthwhile since it
provides a named construct for it, and prevent replicating the code itself
(which would end up in multiple places...)...

PR #351, opened 2022-01-11
Move freeradius container Ubuntu from groovy(20.10) to bionic(20.04)

anurag6:


radius_server/Dockerfile.freeradius
```diff
@@ -1,4 +1,10 @@
-FROM ubuntu:20.10
+FROM ubuntu:20.04
+
+ENV TZ=America/Los_Angeles
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+RUN apt update
```
grafnu, line 6: for whatever reason (as in I forget why) it's customary to put the apt update in the same command (so using &&) -- see the one below

grafnu (approved):

radius_server/Dockerfile.freeradius
anurag6: Missed that. Updated.

PR #352, opened 2022-01-12
Build fix for debian installs

grafnu:


controller/bin/install_target
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash -ex
```
anurag6, line 1: Remove -x? I think that would lead to a bunch of unnecessary prints.

grafnu: done

anurag6 (approved):

